PR #17, opened 2023-03-06
OwnedPool (Chakra) suggested refactor

mtullyoc:
Hi @devjoaov 

I left some suggestions:
- we can use more HStack and VStack rather than Stack. 
- for a sake of simplification, we can trust the default values.

Here we have the same result:
Captura de Tela 2023-03-06 às 15 36 15


devjoaov (approved):
Thanks for the improvements @mtullyoc

PR #18, opened 2023-03-06
Load pools owned from balancer subgraph

devjoaov:
[linear issue](https://linear.app/bleu-llc/issue/BAL-51/load-pools-owned-from-balancer-subgraph)


packages/next-app/src/components/OwnedPool.tsx
```diff
@@ -2,23 +2,46 @@ import { Badge, Button, HStack, Text, VStack } from "@chakra-ui/react";
 
 interface IOwnedPool {
   onClick: () => void;
-  address: string;
-  type: string;
-  ratio?: string;
-  name: string;
+  address?: string;
+  poolType?: string | null | undefined;
```
ribeirojose, line 6: no need to write `poolType?: ... | undefined` -- https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters

packages/next-app/src/components/OwnedPool.tsx
```diff
@@ -2,23 +2,46 @@ import { Badge, Button, HStack, Text, VStack } from "@chakra-ui/react";
...
+  if (!address) return;
+  const match = address.match(
+    /^(0x[a-zA-Z0-9]{10})[a-zA-Z0-9]+([a-zA-Z0-9]{4})$/
+  );
+  if (!match) return address;
+  return `${match[1]}…${match[2]}`;
+}
+
 export function OwnedPool({
   onClick,
```
ribeirojose, line 29: You could also pass only a `pool` prop from the GraphQL request which would essentially be of the same type of a generated type. IMO it's cleaner to read, and you won't need to redefine a bunch of types that had already been defined by the graphql types.

packages/next-app/src/components/OwnedPool.tsx
```diff
@@ -2,23 +2,46 @@ import { Badge, Button, HStack, Text, VStack } from "@chakra-ui/react";
...
+  tokens,
 }: IOwnedPool) {
   const backgroundColor = isSelected ? "blue.800" : "gray.800";
 
+  const poolName =
+    poolType === "Weighted" ? tokens!.map((obj) => obj.symbol).join("/") : name;
+  const weights =
+    poolType === "Weighted"
+      ? tokens!.map((obj) => (Number(obj.weight) * 100).toFixed()).join("/")
+      : null;
```
ribeirojose, line 43: any reason to mark those as a non-null assertion? If there's a scenario where `tokens` is nullish I'd say we need to make the component handle that case, otherwise this could be creating a bug (although it sounds to me more like an impossible scenario caused by the way we defined the component types here).

packages/next-app/src/components/OwnedPool.tsx
```diff
@@ -2,23 +2,46 @@ import { Badge, Button, HStack, Text, VStack } from "@chakra-ui/react";
...
   isSelected,
+  tokens,
 }: IOwnedPool) {
   const backgroundColor = isSelected ? "blue.800" : "gray.800";
 
+  const poolName =
+    poolType === "Weighted" ? tokens!.map((obj) => obj.symbol).join("/") : name;
+  const weights =
+    poolType === "Weighted"
+      ? tokens!.map((obj) => (Number(obj.weight) * 100).toFixed()).join("/")
```
ribeirojose, line 42: you could also do 

```
const formatter = new Intl.NumberFormat("en-US", {
   style: 'percent',
   minimumFractionDigits: 0,
   maximumFractionDigits: 0
});

const weights = tokens.map(({weight}) => formatter.format(weight).replace("%", ""))
```

packages/next-app/src/components/OwnedPool.tsx
```diff
@@ -2,23 +2,46 @@ import { Badge, Button, HStack, Text, VStack } from "@chakra-ui/react";
...
-  ratio,
+  poolType,
   name,
   isSelected,
+  tokens,
 }: IOwnedPool) {
   const backgroundColor = isSelected ? "blue.800" : "gray.800";
 
+  const poolName =
+    poolType === "Weighted" ? tokens!.map((obj) => obj.symbol).join("/") : name;
```
ribeirojose, line 39: Curious: are weighted pool's names not the same as the trading tuple symbols joined by `/`?

packages/next-app/src/lib/gql.ts
```diff
@@ -1,16 +1,42 @@
...
+    case "5":
+      return Network.goerli;
+    case "137":
+      return Network.polygon;
+    case "42161":
+      return Network.arbitrum;
+    default:
+      return Network.mainnet;
+  }
+}
```
ribeirojose, line 29: could also be
```suggestion
function networkFor(key: string | number): Network {
  const networks = {
    "1": Network.mainnet,
    "5": Network.goerli,
    "137": Network.polygon,
    "42161": Network.arbitrum,
  }
  return networks[key.toString()] || Network.mainnet;
}
```

packages/next-app/src/lib/gql.ts
```diff
@@ -1,16 +1,42 @@
...
+} from "@balancer-pool-metadata/balancer-gql/codegen";
+import { getSdkWithHooks as mainnetSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/mainnet";
+import { getSdkWithHooks as goerliSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/goerli";
+import { getSdkWithHooks as polygonSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/polygon";
+import { getSdkWithHooks as arbitrumSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/arbitrum";
 
+const currentNetwork =
+  typeof localStorage !== "undefined"
+    ? localStorage.getItem("networkId") ?? "1"
+    : "1";
```
ribeirojose, line 14: I imagine this value will update if you change the network selector, but it probably won't be read again because this value would've already been evaluated the first time the browser loads this part of the code. Have you tested this?

packages/next-app/src/pages/index.tsx
```diff
@@ -1,15 +1,34 @@
...
-  return (
+  const { chain } = useNetwork();
+  const { isConnected } = useAccount();
+
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    localStorage.setItem("networkId", chain?.id ? String(chain.id) : "1");
+    if (mounted && chain) {
+      Router.reload();
```
ribeirojose, line 17: 👀

packages/next-app/src/lib/gql.ts
```diff
@@ -1,16 +1,42 @@
...
+  [Network.polygon]: polygonSdk,
+  [Network.arbitrum]: arbitrumSdk,
+  [Network.goerli]: goerliSdk,
+};
 
-export default gql
\ No newline at end of file
+const client = new GraphQLClient(ENDPOINTS[networkFor(currentNetwork)]);
+
+const gql = networkSdks[networkFor(currentNetwork)](client);
```
ribeirojose, line 40: if this doesn't work what I'd probably do is wrap the gql client in a function that gets the selected network as a parameter and spits out the instantiated sdk

ribeirojose:
looking good 👍  left a few questions before approving

packages/next-app/src/components/OwnedPool.tsx
devjoaov: The pool names are joined with the token and the weight. I changed it to be just the `token/token/...` and the weights in a bagde component. Here we have an example of the original pool names:
image
Note that I only applied the change to the weighted pools to be closer to our HI-FI

packages/next-app/src/components/OwnedPool.tsx
devjoaov: I forgot to remove it. I'll check if the tokens exist before creating these variables.

packages/next-app/src/lib/gql.ts
devjoaov: When the index page is loaded, we save the `networkId` in the local storage. 

Whenever the `networkId` changes (with the user shifting the network), we refresh the page with `Router.reload()`.

After refreshing the page we set the current networkId on the local storage. As the gqp interface is called after setting the networkId on the local storage, we always have a value for the networkId.

packages/next-app/src/pages/index.tsx
devjoaov: 👆

packages/next-app/src/lib/gql.ts
devjoaov: I don't understand what you mean here, can you elaborate a little more?

ribeirojose: Meant to have something like `const getSdk = (network) => getSdkWithHooks(getClient(network))` but not an issue 👍

PR #20, opened 2023-03-07
Create action to automate deploy

mtullyoc:
[Linear BAL-3](https://linear.app/bleu-llc/issue/BAL-37/create-action-to-automate-deployment-on-push-to-main)


mtullyoc:
@ribeirojose 👀

.github/workflows/graph.yml
```diff
@@ -0,0 +1,35 @@
...
+          pnpm codegen
+          cp subgraph.goerli.yaml subgraph.yaml
+          pnpm build
+
+      - uses: gtaschuk/graph-deploy@v0.1.12
+        with:
+          graph_access_token: ${{secrets.GRAPH_ACCESS_TOKEN}}
+          graph_subgraph_name: "balancer-pools-metadata"
+          graph_account: "bleu-llc"
+          graph_config_file: "./packages/subgraph/subgraph.yaml"
```
mtullyoc, line 33: Should we use `working-directory: ./packages/subgraph` here?

.github/workflows/graph.yml
```diff
@@ -0,0 +1,35 @@
+name: Deploy Graph
+
+on:
+  push:
+    branches: master
```
ribeirojose, line 5:
```suggestion
    branches: main
```
would also add workflow-dispatch to make it easier for us to test it in this PR

.github/workflows/graph.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  deploy-goerli:
+    runs-on: ubuntu-latest
+    environment: graph
+    steps:
+      - uses: actions/checkout@v3
+
```
ribeirojose, line 13: depending on what we do we'd need to recur submodules here

.github/workflows/graph.yml
ribeirojose: Idk, does that work? 😬

packages/subgraph/abis/PoolMetadataRegistry.json
```diff
@@ -1,54 +1,1958 @@
-[
-  {
-    "inputs": [
-      { "internalType": "contract IVault", "name": "vault", "type": "address" }
-    ],
-    "stateMutability": "nonpayable",
-    "type": "constructor"
+{
+  "abi": [
```
ribeirojose, line 2: I'm not sure this is what you want. I think you probably want just what's nested under this `abi` key for this subgraph

ribeirojose (commented):
have you tested this? I'm not comfortable giving an approval without seeing a successful workflow run and the deployed subgraph

.github/workflows/graph.yml
mtullyoc: almost 😬... let's fix the block number

.github/workflows/graph.yml
mtullyoc: I think would be necessary if we needed to run `forge` on contracts package... here we worked on subgraph folder only ✌🏻

.github/workflows/graph.yml
```diff
@@ -0,0 +1,39 @@
+name: Deploy Graph
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - tullyo/bal-37-create-action-to-automate-deployment-on
```
mendesfabio, line 8:
```suggestion
```

mendesfabio (approved):
lgtm!

mendesfabio (approved):

PR #23, opened 2023-03-08
Implement pool metadata view component

devjoaov:
[linear issue](https://linear.app/bleu-llc/issue/BAL-99/implement-pool-metadata-view-component)

Rendered component: 
image

Image modal:

image


packages/next-app/src/app/pool/[poolId]/MetadataAttribute.tsx
```diff
@@ -30,91 +75,114 @@ const Header = () => {
...
+};
 
-  const rows = data.map((row) => [...row, ...defaultActions]);
+const AttributeLink = ({ data }: { data: (typeof cellData)[1] }) => {
+  switch (data.type) {
+    case "URL":
+      return (
+        <button>
+          <ArrowTopRightIcon className="text-white" />
+        </button>
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/MetadataAttribute.tsx:102]: Wouldn't this be a link instead of a button?

packages/next-app/src/app/pool/[poolId]/MetadataAttribute.tsx
```diff
@@ -30,91 +75,114 @@ const Header = () => {
...
+      );
+    case "image":
+      return (
+        <button>
+          <ImageDialog src={data.value as string} />
+        </button>
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/MetadataAttribute.tsx:108]: my personal take is that images should also open on a new tab, like links... but we don't need to spend any time here for now

ribeirojose (approved):
🔝 👌  looking good!

packages/next-app/src/components/ImageDialog.tsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+
+import * as Dialog from "@radix-ui/react-dialog";
+import { ArrowTopRightIcon } from "@radix-ui/react-icons";
+import classNames from "classnames";
```
mendesfabio, line 5: minor thing but you imported as `cn` in other files

ribeirojose: https://github.com/fsubal/eslint-plugin-classnames/blob/main/docs/rules/prefer-classnames-function.md I think we could use this, wdyt @devjoaov ?

PR #27, opened 2023-03-10
Create Pin JSON metadata function

mtullyoc:
[Linear 56](https://linear.app/bleu-llc/issue/BAL-56/create-service-for-pinning-json-file)

https://user-images.githubusercontent.com/17722915/224224542-c262d436-cd3a-42cc-8f1d-cdf5c351140d.mov


packages/next-app/src/app/metadata/[poolId]/page.tsx
```diff
@@ -1,24 +1,82 @@
...
+      "desc": "The address of the smart contract that implements the exchange pool",
+      "value": "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D"
+    },
+    {
+      "name": "Pool link",
+      "type": "URL",
+      "desc": "The address of the smart contract that implements the exchange pool",
+      "value": "https://github.com/"
+    }
+  ]`);
```
Comment (ribeirojose) [inline packages/next-app/src/app/metadata/[poolId]/page.tsx:24]: Couldn't we just JSON.dump this before pinning instead of adding this to a string?

mendesfabio (approved):
just wondering if we need the SDK to use only 1 function from it (and that could be implemented via API) but it's not a big deal to me.

ribeirojose (approved):
don't think any comments are a blocker since we'll need to revisit this when connecting to the overarching submit logic

packages/next-app/src/app/metadata/[poolId]/page.tsx
Comment (mtullyoc) [reply packages/next-app/src/app/metadata/[poolId]/page.tsx:24]: It was getting annoying be swapping between JSON.stringify and JSON.parse on fetch/route handler for this test. Because I got some obscured errors. I didn't spend time on this and let it on string format. 🚀

PR #30, opened 2023-03-13
disable editing typename and key in added metadata

ribeirojose:
- disable key and type editing for existing keys


devjoaov (approved):
Thanks for taking this 👍

PR #31, opened 2023-03-13
Consolidate metadata loading and pinning on the pool metadata page

devjoaov:
[issue 1](https://linear.app/bleu-llc/issue/BAL-122/consolidate-metadata-pinning-on-the-pool[poolid]-page)
[issue 2](https://linear.app/bleu-llc/issue/BAL-120/consolidate-metadata-loading-on-the-pool[poolid]-page)
Fixes BAL-120
Fixes BAL-122


packages/next-app/src/contexts/PoolMetadataContext.tsx
```diff
@@ -50,17 +27,18 @@ interface PoolMetadataContextType {
...
-  // const { data: poolMetadata } = metadataGql.useMetadataPool({
-  //   poolId: selectedPool,
-  // });
-  // cellData = getDataFromIpfs(poolMetadata.CID)
-  const [metadata, setMetadata] = useState<PoolMetadataAttribute[]>(cellData);
+  const [metadata, setMetadata] = useState<PoolMetadataAttribute[]>([]);
+
+  function handleSetMetadata(data: PoolMetadataAttribute[]) {
+    setMetadata(data);
+  }
```
ribeirojose, line 41: useless function definition?

packages/next-app/src/app/pool/[poolId]/page.tsx
```diff
@@ -3,7 +3,7 @@ import { MetadataAttributesTable } from "./(components)/MetadataAttributesTable"
 export default function Page({ params }: { params: { poolId: string } }) {
   return (
     <div className="h-full flex-1 py-5 text-white">
-      <MetadataAttributesTable poolId={params.poolId} />
+      <MetadataAttributesTable poolId={params.poolId as `0x${string}`} />
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/page.tsx:6]: you could also have changed this in the definition in line `3` instead

packages/next-app/src/app/pool/[poolId]/loading.tsx
```diff
@@ -0,0 +1,70 @@
+export default function Loading() {
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/loading.tsx:1]: 🔝

packages/next-app/src/app/pool/[poolId]/loading.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <div className="h-full flex-1 py-5 text-white">
+      <div className="w-full bg-gray-900">
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/loading.tsx:4]: I noticed we have tailwind colors and radix colors 🤔 which do you think we should commit to? No need to go back and refactor everything, but wanted to get your perspective

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
 
-export function MetadataAttributesTable({ poolId }: { poolId: string }) {
-  const { metadata } = useContext(PoolMetadataContext);
+export function MetadataAttributesTable({ poolId }: { poolId: `0x${string}` }) {
+  const { metadata, handleSetMetadata, handleSetPool } =
+    useContext(PoolMetadataContext);
+
+  handleSetPool(poolId);
+
+  useEffect(() => handleSetMetadata([]), []);
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:126]: could you elaborate on what the purpose of this is?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+    async function fetchData() {
+      const response = await fetch(
+        `https://ipfs.io/ipfs/${metadataPool?.metadataCID}`,
+        {
+          method: "GET",
+        }
+      );
+      const json = await response.json();
+      handleSetMetadata(json);
+    }
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:147]: would remove this function definition from the middle of the component

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+  useEffect(() => handleSetMetadata([]), []);
+
+  const { data: pools } = metadataGql.useMetadataPool({
+    poolId,
+  });
+
+  const metadataPool = pools?.pools[0];
+
+  const [poolMetadataCID, setMetadataCID] = useState("");
+  const debouncedCID = useDebounce(poolMetadataCID, 500);
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:135]: any reason to debounce the CID value?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+
+  handleSetPool(poolId);
+
+  useEffect(() => handleSetMetadata([]), []);
+
+  const { data: pools } = metadataGql.useMetadataPool({
+    poolId,
+  });
+
+  const metadataPool = pools?.pools[0];
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:132]:
```suggestion
  const { data } = metadataGql.useMetadataPool({
    poolId,
  });

  const metadataPool = data?.pools?.[0];
```

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+
+  const [poolMetadataCID, setMetadataCID] = useState("");
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:134]: any reason to have this value in a state variable given it's already coming from the GraphQL endpoint?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+  const debouncedCID = useDebounce(poolMetadataCID, 500);
+
+  useEffect(() => {
+    async function fetchData() {
+      const response = await fetch(
+        `https://ipfs.io/ipfs/${metadataPool?.metadataCID}`,
+        {
+          method: "GET",
+        }
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:143]: no need to pass in a [default param ](https://fetch.spec.whatwg.org/#concept-request-method)

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+
+  const metadataPool = pools?.pools[0];
+
+  const [poolMetadataCID, setMetadataCID] = useState("");
+  const debouncedCID = useDebounce(poolMetadataCID, 500);
+
+  useEffect(() => {
+    async function fetchData() {
+      const response = await fetch(
+        `https://ipfs.io/ipfs/${metadataPool?.metadataCID}`,
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:140]: `#question` I know Piñata has an option to make pins public, but I can't remember whether that's set up by default or not -- have you tested this and are we guaranteeing that the things we've pinned are publicly accessible in this endpoint?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+  });
+
+  const metadataPool = pools?.pools[0];
+
+  const [poolMetadataCID, setMetadataCID] = useState("");
+  const debouncedCID = useDebounce(poolMetadataCID, 500);
+
+  useEffect(() => {
+    async function fetchData() {
+      const response = await fetch(
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:139]: what happens if `metadataPool` is empty or undefined or if this request fails?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+        `https://ipfs.io/ipfs/${metadataPool?.metadataCID}`,
+        {
+          method: "GET",
+        }
+      );
+      const json = await response.json();
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:145]: what happens if what you have there is not a `json` file?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
   );
 }
 
-export function MetadataAttributesTable({ poolId }: { poolId: string }) {
-  const { metadata } = useContext(PoolMetadataContext);
+export function MetadataAttributesTable({ poolId }: { poolId: `0x${string}` }) {
+  const { metadata, handleSetMetadata, handleSetPool } =
+    useContext(PoolMetadataContext);
+
+  handleSetPool(poolId);
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:124]: any reasons why the parent wouldn't have already defined this value given it had access to it? Also, bear in mind this is in the URL path so I'm not sure there's a case for having a dedicated state for it.

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+    args: [poolId, debouncedCID],
+    enabled: Boolean(debouncedCID),
+  });
+
+  const { write } = useSetPoolMetadata(config);
+
+  async function pinJSON() {
+    const resp = await fetch(`./${poolId}/pin`, {
+      method: "POST",
+      body: JSON.stringify({ metadata: metadata }),
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:164]:
```suggestion
      body: JSON.stringify({ metadata }),
```

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+      const json = await response.json();
+      handleSetMetadata(json);
+    }
+    if (metadataPool?.metadataCID) {
+      fetchData();
+    }
+  }, [metadataPool?.metadataCID]);
+
+  const { config } = usePrepareSetPoolMetadata({
+    address: "0xebfadf723e077c80f6058dc9c9202bb613de07cf",
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:154]: `#should` live in a separate `constants.json` file.

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+    args: [poolId, debouncedCID],
+    enabled: Boolean(debouncedCID),
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:156]: `#comment` so `debouncedCID` is the new CID that we're going to use to update the contract

`#would` call it something like `updatedMetadataCID` then

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+  const { metadata, handleSetMetadata, handleSetPool } =
+    useContext(PoolMetadataContext);
+
+  handleSetPool(poolId);
+
+  useEffect(() => handleSetMetadata([]), []);
+
+  const { data: pools } = metadataGql.useMetadataPool({
+    poolId,
+  });
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:130]: I'd suggest creating a hook abstracting away the JSON loading logic to make it this component easier to read and expose a simple API that we could eventually publish in a lib.

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+          method: "GET",
+        }
+      );
+      const json = await response.json();
+      handleSetMetadata(json);
+    }
+    if (metadataPool?.metadataCID) {
+      fetchData();
+    }
+  }, [metadataPool?.metadataCID]);
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:151]: `#style` suggestion: return early

```suggestion
    if (!metadataPool?.metadataCID) return
    
    fetchData();
  }, [metadataPool?.metadataCID]);
```

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (mtullyoc) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:140]: [https://ipfs.io/ipfs/QmXJ7yxerhSXQd7DVy2f9CbJ6ieBEGQeiLhDnuRkNL4V8a](https://ipfs.io/ipfs/QmXJ7yxerhSXQd7DVy2f9CbJ6ieBEGQeiLhDnuRkNL4V8a)

My last one is accessible ✅

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+
+  const { config } = usePrepareSetPoolMetadata({
+    address: "0xebfadf723e077c80f6058dc9c9202bb613de07cf",
+    args: [poolId, debouncedCID],
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:155]: `#comment` so `debouncedCID` is the new CID that we're going to use to update the contract

`#would` call it something like `updatedMetadataCID` to make it clearer

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+    enabled: Boolean(debouncedCID),
+  });
+
+  const { write } = useSetPoolMetadata(config);
+
+  async function pinJSON() {
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:161]: good candidate for separating this into a `lib/ipfs.ts`

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
+    }
+  }, [metadataPool?.metadataCID]);
+
+  const { config } = usePrepareSetPoolMetadata({
+    address: "0xebfadf723e077c80f6058dc9c9202bb613de07cf",
+    args: [poolId, debouncedCID],
+    enabled: Boolean(debouncedCID),
+  });
+
+  const { write } = useSetPoolMetadata(config);
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:159]: `#could` rename this (e.g. `writeToMetadataContract`) to make its meaning clearer / stand out in the midst of the component logic

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -111,14 +117,63 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
 
-  function handleUpdatePoolMetadata() {
-    // get `selectedPool` and `metadata` from context
-    // pin metadata to ipfs
-    // call PoolMetadataRegistry contract to setPoolMetadata
-    // TODO: https://linear.app/bleu-llc/issue/BAL-57/update-metadata-calling-poolmetadataregistry-contract
+  async function handleUpdatePoolMetadata() {
+    const pinData = await pinJSON();
+    setMetadataCID(pinData.IpfsHash);
+    write?.();
```
Comment (ribeirojose) [inline packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:176]: should we raise an error when write is nullish?

ribeirojose (changes requested):
In the right direction! Left a few comments, but would take special attention to the `#must`.

ribeirojose:
btw, to link PRs to more than one issue: https://linear.app/docs/github#link-prs

mtullyoc (commented):
Besides José's feedback, LGTM 🎉

IDK whether should be on this PR, but I'd suggest blocking handleUpdatePoolMetadata() unless the metadata has changed. ✌🏻

Another issue I noticed is that on the first click on "Update Metadata" Metamask (write?.()) is not called, although pinJSON() is posted successfully.

packages/next-app/src/contexts/PoolMetadataContext.tsx
devjoaov: From what I know it's generally considered a good practice to pass a function that handles setState through a context, rather than passing the setState function itself

packages/next-app/src/app/pool/[poolId]/page.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/page.tsx:6]: Thanks, I'll apply the change

packages/next-app/src/app/pool/[poolId]/loading.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/loading.tsx:4]: I don't know well the radix colors but seems to be a good option. We can ask about the Nichollas option as well

packages/next-app/src/contexts/PoolMetadataContext.tsx
ribeirojose: oh really, I didn't know that! Curious about the reasoning behind it. Do you have any links you could point me out to?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:126]: the `handleSetPool(poolId);` makes the _selected sidebar pool item_ change the color if we refresh the page. 

and the effect is to avoid the metadata from a pool be shared with a pool that does not have metadata. I'll explain why:

we set the metadata with 

```js
useEffect(() => {
    async function fetchData() {
      const response = await fetch(
        `https://ipfs.io/ipfs/${metadataPool?.metadataCID}`,
        {
          method: "GET",
        }
      );
      const json = await response.json();
      handleSetMetadata(json);
    }
    if (metadataPool?.metadataCID) {
      fetchData();
    }
  }, [metadataPool?.metadataCID]);
```

So if we select a pool with metadata, and then select a pool without metadata, the global state is shared and the pool without metadata gets the metadata from the last pool selected.

To avoid this I created an effect that set the metadata state to an empty array every time that we render the `MetadataAttributesTable` component.

What do you think about that?

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx:140]: In my tests, the [https://ipfs.io/ipfs/](https://ipfs.io/ipfs/) performed better than https://gateway.pinata.cloud/ipfs/

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx]: just removed

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx]: Using swr now

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx]: I'll create a task to handle this

packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx
Comment (devjoaov) [reply packages/next-app/src/app/pool/[poolId]/(components)/MetadataAttributesTable.tsx]: agreed

ribeirojose (approved):

PR #35, opened 2023-03-16
Create function to check whether the DAO owns given pool

devjoaov:
fixes [BAL-71](https://linear.app/bleu-llc/issue/BAL-71/create-function-to-check-whether-the-dao-owns-given-pool)

### DIFF
--- a/packages/contracts/src/PoolMetadataAuthorization.sol
+++ b/packages/contracts/src/PoolMetadataAuthorization.sol
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.0 <0.9.0;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IAuthorizer.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
+import {BasePoolAuthorization} from "balancer-v2-monorepo/pkg/pool-utils/contracts/BasePoolAuthorization.sol";
+
+abstract contract PoolMetadataAuthorization is Authentication {
+    address internal constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
+
+    /// @notice Returns the authorizer contract for managing access control
+    /// @return The authorizer contract
+    function getAuthorizer() external view returns (IAuthorizer) {
+        return _getAuthorizer();
+    }
+
+    /// @notice Determines whether an action can be performed by a given account
+    /// @param actionId The ID of the action to be performed
+    /// @param account The account to check for authorization
+    /// @return bool Returns true if the action can be performed, otherwise false
+    function _canPerform(bytes32 actionId, address account) internal view override returns (bool) {
+        if (_getOwner() != _DELEGATE_OWNER) {
+            // Only the owner can perform "owner only" actions, unless the owner is delegated.
+            return msg.sender == _getOwner();
+        } else {
+            // Non-owner actions are always processed via the Authorizer, as "owner only" ones are when delegated.
+            return _getAuthorizer().canPerform(actionId, account, address(this));
+        }
+    }
+
+    /// @notice Returns the owner of the pool that's currently being operated on
+    /// @return The owner of the pool
+    function _getOwner() internal view virtual returns (address);
+
+    function _getAuthorizer() internal view virtual returns (IAuthorizer);
+}
--- a/packages/contracts/src/PoolMetadataAuthorization.sol
+++ b/packages/contracts/src/PoolMetadataAuthorization.sol
@@ -8,12 +8,6 @@ import {BasePoolAuthorization} from "balancer-v2-monorepo/pkg/pool-utils/contrac
 abstract contract PoolMetadataAuthorization is Authentication {
     address internal constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
 
-    /// @notice Returns the authorizer contract for managing access control
-    /// @return The authorizer contract
-    function getAuthorizer() external view returns (IAuthorizer) {
-        return _getAuthorizer();
-    }
-
     /// @notice Determines whether an action can be performed by a given account
     /// @param actionId The ID of the action to be performed
     /// @param account The account to check for authorization
@@ -32,5 +26,7 @@ abstract contract PoolMetadataAuthorization is Authentication {
     /// @return The owner of the pool
     function _getOwner() internal view virtual returns (address);
 
+    /// @notice Returns the authorizer contract for managing access control
+    /// @return The authorizer contract
     function _getAuthorizer() internal view virtual returns (IAuthorizer);
 }
--- a/packages/contracts/src/PoolMetadataAuthorization.sol
+++ b/packages/contracts/src/PoolMetadataAuthorization.sol
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.7.0 <0.9.0;
 
-import "@balancer-labs/v2-interfaces/contracts/vault/IAuthorizer.sol";
-import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
+import "balancer-v2-monorepo/pkg/interfaces/contracts/vault/IAuthorizer.sol";
+import "balancer-v2-monorepo/pkg/solidity-utils/contracts/helpers/Authentication.sol";
 import {BasePoolAuthorization} from "balancer-v2-monorepo/pkg/pool-utils/contracts/BasePoolAuthorization.sol";
 
 abstract contract PoolMetadataAuthorization is Authentication {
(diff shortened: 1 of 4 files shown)


packages/contracts/src/PoolMetadataAuthorization.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.0 <0.9.0;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IAuthorizer.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
```
ribeirojose, line 5: `#would` use only imports from `balancer-v2-monorepo/pkg`. Importing from @balancer-labs will require installing npm packages and linking them through remappings which we proved at the beginning of the project is super error prone

packages/contracts/test/PoolMetadataRegistry.t.sol
```diff
@@ -52,7 +58,28 @@ contract PoolMetadataRegistryTest is IPoolMetadataRegistry, Test {
...
+            1e16,
+            0,
+            0,
+            _DELEGATE_OWNER
+        );
+
         _poolMetadataRegistry = new MockPoolMetadataRegistry(_vault);
+
+        // Compute the "setPoolMetadata(bytes32,string)" actionId
+        bytes32 _actionIdDisambiguator = bytes32(uint256(address(_poolMetadataRegistry)));
```
mendesfabio, line 77: only private vars should start with underscore
```suggestion
        bytes32 actionIdDisambiguator = bytes32(uint256(address(_poolMetadataRegistry)));
```

packages/contracts/test/PoolMetadataRegistry.t.sol
```diff
@@ -14,22 +17,25 @@ contract MockPoolMetadataRegistry is PoolMetadataRegistry {
...
+    MockBasicAuthorizer private _authorizer;
+    MockAuthenticatedContract private _authenticated;
     IVault private _vault;
     MockBasePool private _basePool;
+    MockBasePool private _delegatedBasePool;
     address private constant _poolOwner = address(15);
 
     string private constant _testMetadataCID = "QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR";
+    address internal constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
+    address DAOmultisig = 0x10A19e7eE7d7F8a52822f6817de8ea18204F2e4f;
```
mendesfabio, line 33: I don't think this needs to be a real address - in this case, you're using the DAO multi-sig on Ethereum, but I'd rather generate a random one to avoid confusion.

packages/contracts/test/PoolMetadataRegistry.t.sol
```diff
@@ -98,6 +115,23 @@ contract PoolMetadataRegistryTest is IPoolMetadataRegistry, Test {
...
+        _poolMetadataRegistry.setPoolMetadata(_delegatedBasePool.getPoolId(), _testMetadataCID);
+
+        assertEq(_poolMetadataRegistry.poolIdMetadataCIDMap(_delegatedBasePool.getPoolId()), _testMetadataCID);
+        vm.stopPrank();
+    }
+
+    function testMetadataSetterWhenDAONotDelegatedOwner() public {
+        bytes32 poolId = _basePool.getPoolId();
+        vm.startPrank(DAOmultisig);
+        vm.expectRevert("BAL#401");
```
mendesfabio, line 129: worth commenting about what this error BAL#401 means.

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -12,15 +16,21 @@ interface IPoolMetadataRegistry {
 /// @title A Pool Metadata dApp
 /// @author Bleu LLC
 /// @notice This contract provides a registry for Balancer's Pools metadata
-contract PoolMetadataRegistry is IPoolMetadataRegistry {
+contract PoolMetadataRegistry is IPoolMetadataRegistry, PoolMetadataAuthorization {
     IVault private immutable _vault;
-
+    bytes32 private _poolId;
```
mendesfabio, line 21: hmm, it sounds confusing to me declaring this `_poolId` -- IIRC it's required for `_getOwner` function?

packages/contracts/src/PoolMetadataAuthorization.sol
```diff
@@ -0,0 +1,36 @@
...
+            return msg.sender == _getOwner();
+        } else {
+            // Non-owner actions are always processed via the Authorizer, as "owner only" ones are when delegated.
+            return _getAuthorizer().canPerform(actionId, account, address(this));
+        }
+    }
+
+    /// @notice Returns the owner of the pool that's currently being operated on
+    /// @return The owner of the pool
+    function _getOwner() internal view virtual returns (address);
```
mendesfabio, line 33: couldn't you re-write this to receive the pool address as an argument?

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -12,15 +16,21 @@ interface IPoolMetadataRegistry {
...
     mapping(bytes32 => string) public poolIdMetadataCIDMap;
 
-    constructor(IVault vault) {
+    constructor(IVault vault) Authentication(bytes32(uint256(address(this)))) PoolMetadataAuthorization() {
         _vault = vault;
     }
 
+    modifier onlyRegisteredPool(bytes32 poolId) {
+        _poolId = poolId;
```
ribeirojose, line 29: Any reason to set this value here? I wouldn't expect this function to have side effects in the instance variables.

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -12,15 +16,21 @@ interface IPoolMetadataRegistry {
...
+        require(_isPoolRegistered(poolId), "Pool not registered");
+        _;
```
ribeirojose, line 31: TIL: https://docs.soliditylang.org/en/develop/contracts.html#function-modifiers

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -32,21 +42,27 @@ contract PoolMetadataRegistry is IPoolMetadataRegistry {
...
     /// @notice Updates the pool metadata CID
     /// @param poolId The pool ID to update the metadata
     /// @param metadataCID The metadataCID related to the new pool metadata
-    function setPoolMetadata(bytes32 poolId, string memory metadataCID) public {
-        require(_isPoolRegistered(poolId), "Pool not registered");
-        require(_isPoolOwner(poolId), "Caller is not the owner");
+    function setPoolMetadata(bytes32 poolId, string memory metadataCID)
+        public
+        onlyRegisteredPool(poolId)
+        authenticate
```
ribeirojose, line 64: I assume this modifier is defined in `Authentication` and uses your `_getOwner` function, right? Since there's no way to know that wo knowing the context can you add a brief explanation of how that works here?

packages/contracts/src/PoolMetadataAuthorization.sol
devjoaov: No, I've tried a few ways but to no avail. The solution I found was to create a variable and update its value before running the authenticate modifier.

packages/contracts/src/PoolMetadataRegistry.sol
devjoaov: yes, to get the pool owner I need the poolId. 

1. The authenticate modifier calls the canPerform. 
2. the canPerform just receive the `bytes32 actionId` and the `address account`
3. the canPerform calls getOwner without any arguments

Any idea of how to solve it with another approach?

packages/contracts/src/PoolMetadataRegistry.sol
devjoaov:
```js
function setPoolMetadata(bytes32 poolId, string memory metadataCID)
        public
        onlyRegisteredPool(poolId)
        authenticate
    {
        poolIdMetadataCIDMap[poolId] = metadataCID;
        emit PoolMetadataUpdated(poolId, metadataCID);
    }
```
as I said above to @mendesfabio, we need the poolId inside the getOwner (a function that does not receive parameters)

1. The authenticate modifier calls the `canPerform`.
2. the `canPerform` receive the `bytes32 actionId` and the `address account`
3. the `canPerform` calls `getOwner` without any arguments

As we need the poolId, I created a variable and we set its value before performing the authenticate modifier on the setPoolMetadata.

packages/contracts/src/PoolMetadataAuthorization.sol
```diff
@@ -0,0 +1,36 @@
...
+import {BasePoolAuthorization} from "balancer-v2-monorepo/pkg/pool-utils/contracts/BasePoolAuthorization.sol";
+
+abstract contract PoolMetadataAuthorization is Authentication {
+    address internal constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
+
+    /// @notice Returns the authorizer contract for managing access control
+    /// @return The authorizer contract
+    function getAuthorizer() external view returns (IAuthorizer) {
+        return _getAuthorizer();
+    }
```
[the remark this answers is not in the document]
devjoaov: No, I've just removed it.

PR #36, opened 2023-03-17
Show DAO's pool list using Multisig addresses

mtullyoc:
Closes [BAL-63](https://linear.app/bleu-llc/issue/BAL-63/list-pools-owned-by-the-dao-when-a-signer-connects)

https://www.loom.com/share/f8208b4fb9f54217998b1573cb2e1083

### DIFF
--- a/packages/next-app/src/lib/gql.ts
+++ b/packages/next-app/src/lib/gql.ts
@@ -47,11 +47,12 @@ const client = new GraphQLClient(ENDPOINTS[networkFor(currentNetwork)]);
 const gql = networkSdks[networkFor(currentNetwork)](client);
 
 export function impersonateWhetherDAO(address: `0x${string}` | undefined) {
-  if (networkMultisigs[networkFor(currentNetwork)] === (address as string)) {
-    return DELEGATE_OWNER
-  }
+  const network = networkFor(currentNetwork);
 
-  return address
+  if (network !== Network.goerli && networkMultisigs[network] === address) {
+    return DELEGATE_OWNER;
+  }
+  return address;
 }
 
 export default gql;
--- a/packages/next-app/src/lib/gql.ts
+++ b/packages/next-app/src/lib/gql.ts
@@ -1,6 +1,6 @@
 import {
   ENDPOINTS,
-  Network,
+  Network
 } from "@balancer-pool-metadata/balancer-gql/codegen";
 import { getSdkWithHooks as arbitrumSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/arbitrum";
 import { getSdkWithHooks as goerliSdk } from "@balancer-pool-metadata/balancer-gql/src/gql/__generated__/goerli";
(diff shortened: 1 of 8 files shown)


packages/next-app/src/lib/gql.ts
```diff
@@ -34,8 +36,47 @@ const networkSdks = {
...
+  '0xc38c5f97B34E175FFd35407fc91a937300E33860',
+  '0xeE071f4B516F69a1603dA393CdE8e76C40E5Be85',
+]
+
+const networkMultisig: {
+  [key: string]: string[]
+} = {
+  [Network.mainnet]: mainnetMultisig,
+  [Network.polygon]: polygonMultisig,
+  [Network.arbitrum]: arbitrumMultisig,
```
ribeirojose, line 67: We'd still want to add Optimism here right?

packages/pool-metadata-gql/codegen.ts
```diff
@@ -3,16 +3,17 @@ import { CodegenConfig } from "@graphql-codegen/cli";
...
 
+// TODO: https://linear.app/bleu-llc/issue/BAL-131/deploy-contracts-in-all-networks-that-balancer-is-deployed
 export const ENDPOINTS = {
-  // [Network.mainnet]: `${BASE_URL}/balancer-v2`,
-  // [Network.polygon]: `${BASE_URL}/balancer-polygon-v2`,
-  // [Network.arbitrum]: `${BASE_URL}/balancer-arbitrum-v2`,
+  [Network.mainnet]: `${BASE_URL}/balancer-pools-metadata`,
+  [Network.polygon]: `${BASE_URL}/balancer-pools-metadata`,
+  [Network.arbitrum]: `${BASE_URL}/balancer-pools-metadata`,
   [Network.goerli]: `${BASE_URL}/balancer-pools-metadata`,
```
ribeirojose, line 17: ok, so we added these endpoints here as an example but we're not fully leveraging them 👍 sgtm. Could we maybe then just add the full list of networks we're going to need to deploy to? It could use the same endpoints etc.

packages/next-app/src/lib/gql.ts
```diff
@@ -34,8 +36,47 @@ const networkSdks = {
...
+  '0x7c68c42De679ffB0f16216154C996C354cF1161B',
+  '0xc38c5f97B34E175FFd35407fc91a937300E33860',
+  '0xaF23DC5983230E9eEAf93280e312e57539D098D0',
+]
+
+const polygonMultisig = [
+  '0x7c68c42De679ffB0f16216154C996C354cF1161B',
+  '0xc38c5f97B34E175FFd35407fc91a937300E33860',
+  '0xeE071f4B516F69a1603dA393CdE8e76C40E5Be85',
+]
```
ribeirojose, line 60: not sure if we would be hardcoding all addresses here, would ask @mendesfabio, and I also know some of them have DAO members and some the Maxi as signers. I'll also ask the grants folks directly

packages/next-app/src/lib/gql.ts
mendesfabio: Balancer just launched on Gnosis and Avax should happen soonTM. we should support both.

packages/next-app/src/lib/gql.ts
mendesfabio: most of these addresses are related to other DAO operations. we only care about the ones that have permission to set swap fees: Mainnet Fee Setter (Eth) and Protocol Fees Multisig (Arb and Poly)

image

packages/next-app/src/lib/gql.ts
mendesfabio: On Gnosis maybe the permission wasn't even granted by the DAO since it just launched.

PR #37, opened 2023-03-21
Implement missing pages

devjoaov:
fixes BAL-153
fixes BAL-133
[linear issue](https://linear.app/bleu-llc/issue/BAL-133/remove-load-more-from-pool-page)

printscreens: 
image
image
image


packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx
```diff
@@ -92,7 +94,9 @@ export function TransactionDialog({
     handleSubmit,
   } = useContext(PoolMetadataContext);
 
-  const { data: poolsData } = metadataGql.useMetadataPool({
+  const { data: poolsData } = metadataGql(
+    chain?.id.toString() || "1"
```
Comment (mendesfabio) [inline packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx:98]: why should we fallback to Ethereum here?

Comment (devjoaov) [reply packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx:98]: the chain can be undefined, in our case, it won't be undefined since we just render this component for connected wallets. 
it's just to avoid typescript errors

Comment (mendesfabio) [reply packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx:98]: if you're sure it won't be undefined then the post-fix operator should be a better approach: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

packages/next-app/src/app/metadata/(components)/RootLayout.tsx
```diff
@@ -23,18 +24,19 @@ import { OwnedPool } from "./OwnedPool";
...
-    router.refresh();
-  }, [chain]);
+  address = impersonateWhetherDAO(chain?.id.toString() || "1", address);
+
+  const { data } = gql(chain?.id.toString() || "1").usePool({
+    owner: address,
+  });
 
   return (
     <WagmiConfig client={client}>
```
ribeirojose, line 42: Wonder if we should split this layout so that it could be reused by the admin app.

packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx
Comment (ribeirojose) [reply packages/next-app/src/app/metadata/pool/[poolId]/(components)/TransactionDialog.tsx:98]: You can also use `tiny-invariant`
```
import invariant from 'tiny-invariant';

invariant(chain.id, 'Chain empty :E');
```

ribeirojose (approved):

PR #43, opened 2023-03-21
[Pool Metadata] Disable updating metadata button if the metadata has not changed

devjoaov:
[linear issue](https://linear.app/bleu-llc/issue/BAL-127/disable-updating-metadata-button-if-the-metadata-was-not-changed)

disabled button and link to balancer pool printscreen

image


packages/next-app/src/app/metadata/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -109,15 +112,29 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
...
       <div className="pr-4 sm:pr-6 lg:pr-12">
         <div className="sm:flex sm:items-center">
           <div className="sm:flex-auto">
-            <h1 className="mx-1 text-2xl font-medium text-gray-400">
-              Metadata attributes - Pool #{poolId}
+            <h1 className="mx-1 flex text-2xl font-medium text-gray-400">
+              Metadata attributes - Pool
+              <a
+                target="_blank"
+                href={link}
```
Comment (mendesfabio) [inline packages/next-app/src/app/metadata/pool/[poolId]/(components)/MetadataAttributesTable.tsx:132]: could you rename link explicitly say it's a link to the Balancer app?

packages/next-app/src/app/metadata/pool/[poolId]/(components)/MetadataAttributesTable.tsx
```diff
@@ -109,15 +112,29 @@ function Row({ data }: { data: PoolMetadataAttribute }) {
 }
 
 export function MetadataAttributesTable({ poolId }: { poolId: `0x${string}` }) {
-  const { metadata, handleSubmit } = useContext(PoolMetadataContext);
+  const { metadata, handleSubmit, metadataUpdated } =
+    useContext(PoolMetadataContext);
+
+  const { chain } = useNetwork();
+  const link = `https://app.balancer.fi/#/${networkFor(
+    chain!.id.toString()
```
Comment (mendesfabio) [inline packages/next-app/src/app/metadata/pool/[poolId]/(components)/MetadataAttributesTable.tsx:120]: 😎

packages/next-app/src/contexts/PoolMetadataContext.tsx
```diff
@@ -65,16 +66,20 @@ export function PoolMetadataProvider({ children }: { children: ReactNode }) {
     UpdateStatus.PINNING
   );
   const [submit, handleSubmit] = useState<boolean>(false);
+  const [metadataUpdated, setMetadataUpdated] = useState<boolean>(false);
```
mendesfabio, line 69: for instance, if the user changes one field, and then changes it back to the previous value (i.e. the one currently on IPFS) -- is `metadataUpdated` going to be true? 

I don't think that's a problem for now but we should log all improvements that still can be made on Linear

devjoaov: Thanks for the catch!

I've updated the context to compare the original metadata state with the updated metadata. Now if we change a field and roll back to the same value, the update button will be disabled.

packages/next-app/src/contexts/PoolMetadataContext.tsx
```diff
@@ -1,11 +1,12 @@
 "use client";
-
 import { TypenameEnum } from "@balancer-pool-metadata/schema";
+import { isEqual } from "lodash";
```
ribeirojose, line 3: I'm not sure where tree-shaking is at this point, and I think lodash is a commonjs module and that webpack can only do tree shaking with ES6 modules(to be confirmed, but that's the last I remember), so to be safe would only import to avoid bloating the bundle size

```suggestion
import isEqual from "lodash/isEqual";
```

devjoaov: I'll research about that, thanks 👍

PR #47, opened 2023-03-24
[DAO Admin] Input to filter by Operation Responsible

luizakp:
[Linear issue](https://linear.app/bleu-llc/issue/BAL-163/input-to-filter-by-responsible)

https://www.loom.com/share/47c0feb711ee4a88b86e28a01a2e3c68


packages/next-app/package.json
```diff
@@ -22,7 +22,8 @@
     "@radix-ui/colors": "^0.1.8",
     "@radix-ui/react-alert-dialog": "^1.0.3",
     "@radix-ui/react-dialog": "^1.0.3",
-    "@radix-ui/react-icons": "^1.3.0",
+    "@radix-ui/react-dropdown-menu": "^2.0.4",
```
ribeirojose, line 25: Why do a dropdown menu instead of a select?

packages/next-app/src/contexts/AdminToolsContext.tsx
```diff
@@ -55,15 +56,43 @@ interface AdminToolsContextType {
...
+  selectedFilters: { [key: string]: string };
+  changeSelectedFilters: (key: string, value: string) => void;
+  clearSelectedFilter: (field: string) => void;
 }
 
 export const AdminToolsContext = createContext({} as AdminToolsContextType);
 
 export function AdminToolsProvider({ children }: { children: ReactNode }) {
+  const filterInitialState = {
+    "operation-responsible": "",
```
ribeirojose, line 68: any reason for this to be a kebab case?

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,85 @@
+import { MagnifyingGlassIcon } from "@radix-ui/react-icons";
+import Link from "next/link";
+import { useState } from "react";
```
ribeirojose, line 3: I have a feeling that if this is to live under `/components` its API must not be tied to your use case in daoadmin, otherwise it's just gonna be impossible to assume that these components can be used by any app and/or whomever wants to reuse a component will have to spend a lot of time refactoring it and making sure it doesn't break anything in the original implementation. WDYT?

packages/next-app/src/components/FilterDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </button>
+      </DropdownMenu.Trigger>
+
+      <DropdownMenu.Portal>
+        <DropdownMenu.Content
+          sideOffset={5}
+          className="rounded-md bg-white px-5 py-2 will-change-[opacity,transform] data-[side=top]:animate-slideDownAndFade data-[side=right]:animate-slideLeftAndFade data-[side=bottom]:animate-slideUpAndFade data-[side=left]:animate-slideRightAndFade"
+        >
+          {filters.map((filter) => {
+            const [filterValue, setFilterValue] = useState("");
```
ribeirojose, line 38: This should've triggered an eslint error, but this breaks one of the rules of hooks and could lead to bugs, see [this](https://legacy.reactjs.org/docs/hooks-rules.html)

ribeirojose (approved):
To me the only `#musts` are 
- refactoring the filter dropdown to not depend on the dao admin context, and 
- refactor the nested loop return value to be an actual component so that you can use useState within it wo potentially generating bugs.

Aside from that, I don't fully grasp why not simply using a select for the filter. If we're adding more complex filtering I'd say we should probably revisit the UI design. I'll the leave the rest as recommendations, but would really consider combining components that are commons to both apps into the `components/` folder to avoid them drifting too much from one another and also so that we're able to quickly make edits and add new mini-apps like this to our suite. 

Also one cool #idea (at least imo lol) would be adding a [command menu](https://cmdk.paco.me/) to automatically add filters/parse through.

packages/next-app/src/components/Sidebar.tsx
luizakp: Now looking I was thinking the same, for now I'll move to `/daoadmin` and think on how to implement as a component for all apps later on

packages/next-app/package.json
ribeirojose: in that case I think we should aim at having multiple selects, each for each filter, like Google does, and not complicate our life with a very custom component.

packages/next-app/src/components/Sidebar.tsx
ribeirojose: defer to you, but I don't see us being in a hurry such that taking more time here would lead to a tradeoff in delaying our delivery time since we don't have a hard delivery timeline, so would really optimize for learning and code quality until we do (and also because, when we do, these shortcuts will _also_ cost us time to deliver new things and adapt what we've done)

packages/next-app/package.json
luizakp: created an [issue here](https://linear.app/bleu-llc/issue/BAL-178/multiple-selects-instead-of-dropdown)

PR #48, opened 2023-03-27
[DAO Admin] Verify pool ID given by user and display symbol

luizakp:
[Linear Issue](https://linear.app/bleu-llc/issue/BAL-164/verify-poolid-given-by-user-and-display-symbol)

For this, I implemented an useEffect to check the input `poolID` length, if the length is 66 then a fetch to the graphql is made

https://www.loom.com/share/09d865bd5bd94fc6b643fc9bac400eb7


packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,83 @@
...
+        fetchExistingPool(poolID, chain?.id.toString() || "1").then(
+          (response) => setPoolName(response.pool?.symbol as string)
+        );
+      else {
+        setPoolName("");
+        setPoolError("Pool not found. Please insert an existing Pool ID");
+      }
+    } else {
+      setPoolName("");
+      setPoolError("");
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:78]: you could just use the generated the generated `usePoolExists` swr hook generated by your adding the graphql query

packages/balancer-gql/src/queries/pools.ts
```diff
@@ -14,3 +14,9 @@ export const poolWhereOwner = gql`
     }
   }
 `;
+
+export const poolExists = gql`
+  query PoolExists($poolAddress: ID!) {
+    pool(id: $poolAddress) { symbol }
+  }
+`
```
ribeirojose, line 22: imo this name gives the impression you'll get a boolean in return, which is incorrect. Would rename it to something more accurate like `Pool` and then use the generated `usePool` hook

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,83 @@
...
+      ) : (
+        <></>
+      )}
+    </div>
+  );
+};
+
 export function ActionAttributeContent() {
-  const { register, handleSubmit } = useForm();
+  const { register, handleSubmit, watch } = useForm();
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:55]: could you say more to what made you prefer using a form here?

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -76,12 +122,18 @@ export function ActionAttributeContent() {
...
+                      <Input
+                        label={field.name}
+                        placeholder={field.placeholder}
+                        {...register(toCammelCase(`${field.name}`))}
+                      />
+                      <CheckPoolID
+                        fieldName={field.name}
+                        poolName={poolName}
+                        poolError={poolError}
+                      />
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:135]: any reason to abstract this into a component? I don't see it's abstracting any functionality that other components could use going up and down to check conditions made the code harder to reason about imo

packages/next-app/src/utils/fetcher.ts
```diff
@@ -15,3 +15,11 @@ export async function fetchOwnedPools(address: string, chainId: string) {
...
   return data;
 }
+
+export async function fetchExistingPool(poolAddress:string ,chainId: string) {
+  if (!poolAddress) return { pool: {} };
+  const data = await gql(chainId).PoolExists({
+    poolAddress: poolAddress,
+  });
+  return data;
+}
```
ribeirojose, line 25: do you see a case for calling the query directly instead of using the generated hook? `#should` use the generated hook imo.

packages/next-app/src/contexts/AdminToolsContext.tsx
```diff
@@ -17,6 +18,17 @@ export function toSlug(string?: string) {
   );
 }
 
+export function toCammelCase(string: string) {
```
ribeirojose, line 21:
```suggestion
export function toCamelCase(string: string) {
```

packages/next-app/src/contexts/AdminToolsContext.tsx
```diff
@@ -55,15 +67,43 @@ interface AdminToolsContextType {
...
+
+  function clearSelectedFilter(field: string) {
+    setSelectedFilters((prevState) => {
+      return {
+        ...prevState,
+        [field]: "",
+      };
+    });
+  }
+
```
ribeirojose, line 106: `#could` make these functions API more clear. I think `key` in `changeSelectedFilters` and `field` in `clearSelectedFilters` refer to the same thing, right? So they should probably have the same name. Also, you could leverage Typescript to qualify  a bit more what the allowed filterKeys are and instead of saying that they're a string you can type these params with the allowed filterKeys from a preset of values using e.g. an Enum

packages/next-app/src/contexts/AdminToolsContext.tsx
ribeirojose: Also this is the sort of thing that's so abstract it should be abstracted away from the component code to avoid other folks reimplementing the same thing.

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,83 @@
...
+        <>
+          <h1 className="text-white">Pool Name:</h1>
+          <h1>{poolName}</h1>
+        </>
+      ) : fieldName === "Pool ID" && poolError ? (
+        <h1>{poolError}</h1>
+      ) : (
+        <></>
+      )}
+    </div>
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:50]: This generated more lines of code to read than if it was just part of the component, I'm afraid 😬

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,83 @@
...
-            ref={ref}
-            {...rest}
-            className="selection:color-white box-border inline-flex h-[35px] w-full appearance-none items-center justify-center rounded-[4px] bg-blackA5 px-[10px] text-[15px] leading-none text-white shadow-[0_0_0_1px] shadow-blackA9 outline-none selection:bg-blackA9 hover:shadow-[0_0_0_1px_black] focus:shadow-[0_0_0_2px_black] disabled:bg-blackA9"
-          />
-        </div>
-      );
+  const poolID = watch("poolID");
+
+  React.useEffect(() => {
+    if (poolID) {
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:67]: in this case you can return early instead of doing this do avoid more indents to your code and requiring the reader to keep the whole control flow context in mind:

```suggestion
    if (!ethers.utils.isAddress(poolId)) return;
```

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:78]: You could just initialize these values in those states to avoid needing them here.

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -76,12 +122,18 @@ export function ActionAttributeContent() {
...
-                      key={field.name}
-                      label={field.name}
-                      placeholder={field.placeholder}
-                      {...register(toSlug(`${field.name}`))}
-                    />
+                    <div key={field.name}>
+                      <Input
+                        label={field.name}
+                        placeholder={field.placeholder}
+                        {...register(toCammelCase(`${field.name}`))}
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:129]: Since you're using `react-input-form` you could already leverage their validations to validate that the input value is a valid Pool address.

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { FilterDropdown, IFilter } from "./FilterDropdown";
+
+export function Sidebar() {
+  const data = hardcodedData;
+  const [querySearch, setQuerySearch] = useState("");
+  const { selectedAction, handleSetAction, selectedFilters } = useAdminTools();
+
+  const handleSearchChange = (e: React.FormEvent<HTMLInputElement>) => {
+    setQuerySearch(e.currentTarget.value);
+  };
```
ribeirojose, line 18: Is there any reason to have this living here instead of centralizing this filtering related logic to the Admin context? 

IMO this absolutely must not live here, and I think it only strengthens the case of creating a separate context for filtering. Also `#would` consider if there's any reason to put the search query on the URL here. What's the value of putting the search query in the URL? Would users share this URL with other folks? If not, would consider not having this at all to avoid complicating things.

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const filteredActionsByName = filteredActionsByTag.filter((item) =>
+    item.name.toLowerCase().includes(querySearch.toLowerCase())
+  );
+
+  const filters: IFilter[] = [
+    {
+      name: "Operation Responsible",
+      options: ["Maxis", "Labs"],
+    },
+  ];
```
ribeirojose, line 37: `#must` refactor this out of the Sidebar component. Also, just as a `meta` comment but I think this is more a list of items (ActionListMenu, if you will) than can be called a Sidebar which is usually alongside the main display area and not conditional on the number of items that are in it. 

Also a very strongly `#would` imo is refactoring this to avoid code duplication between the metadata and daoadmin. Even more considering this is placed in the root `components` folder but it has a bunch of daoadmin logic intertwined.

packages/next-app/src/components/FilterDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <DropdownMenu.Content
+          sideOffset={5}
+          className="rounded-md bg-white px-5 py-2 will-change-[opacity,transform] data-[side=top]:animate-slideDownAndFade data-[side=right]:animate-slideLeftAndFade data-[side=bottom]:animate-slideUpAndFade data-[side=left]:animate-slideRightAndFade"
+        >
+          {filters.map((filter) => {
+            const [filterValue, setFilterValue] = useState("");
+
+            function handleSelectFilter(value: string) {
+              setFilterValue(value);
+              changeSelectedFilters(toCammelCase(filter.name), value);
```
ribeirojose, line 42: do we need different cases for any reason? This is super error prone in general but needed especially when interfacing between different code conventions, which doesn't seem to be the case here (plz correct me if I'm wrong). If that's the case `#should` avoid needing to change this to camelCase at first and just have `filter.name` as what should be the parameter.

ribeirojose (changes requested):
left a few suggestions and some strong `#musts`. I feel like I might've repeated myself here -- perhaps we lost the context of the other discussion we were having?

ribeirojose:
Ah ok, this is pointed to `main` directly, now I see why I feel like I repeated myself lol. @luizakp can you take a look at that?

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
Comment (luizakp) [reply packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:78]: use the` usePool` instead of the `fetchExistingPool`?

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,83 @@
...
-            {...rest}
-            className="selection:color-white box-border inline-flex h-[35px] w-full appearance-none items-center justify-center rounded-[4px] bg-blackA5 px-[10px] text-[15px] leading-none text-white shadow-[0_0_0_1px] shadow-blackA9 outline-none selection:bg-blackA9 hover:shadow-[0_0_0_1px_black] focus:shadow-[0_0_0_2px_black] disabled:bg-blackA9"
-          />
-        </div>
-      );
+  const poolID = watch("poolID");
+
+  React.useEffect(() => {
+    if (poolID) {
+      if (poolID.length === 66)
```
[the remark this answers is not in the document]
Comment (luizakp) [reply packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:53]: great! My thought process was just to not call the request on every onChange, just when the input value could really be a pool ID. I did this because I've faced timeout probably due to too many requests in a short period of time

packages/next-app/src/contexts/AdminToolsContext.tsx
```diff
@@ -55,15 +67,43 @@ interface AdminToolsContextType {
   handleSetAction: (action: ActionAttribute) => void;
   submit: boolean;
   handleSubmit: Dispatch<SetStateAction<boolean>>;
+  selectedFilters: { [key: string]: string };
+  changeSelectedFilters: (key: string, value: string) => void;
+  clearSelectedFilter: (field: string) => void;
```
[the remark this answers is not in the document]
luizakp: already as a state from #47 feedback

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
Comment (luizakp) [reply packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:40]: 🤔 this is not for the initial state, it is for when the user clears the input completely after typing something

packages/next-app/src/components/Sidebar.tsx
luizakp: Already removed from components 😃 

Not quite sure what you mean by 'search query on the URL'. This is just a search query to navigate through the DAO Actions. Could you clarify what you meant? 🙂

packages/next-app/src/contexts/AdminToolsContext.tsx
luizakp: `#could `part is done, thinking on the allowed filterKeys part

packages/next-app/src/components/FilterDropdown.tsx
luizakp: the `filter.name` is for example `Operation Responsible` so with the camel case transformation I can filter all the actions with 

``` js

data.actions.filter(
          (item) =>
            item.operationResponsible === selectedFilters.operationResponsible
        )
```

luizakp: I think with the [enum suggestion ](https://github.com/bleu-studio/balancer-pool-metadata/pull/48#discussion_r1149283728) this can be solved

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,37 +1,45 @@
...
+  const poolId = watch("poolID");
+
+  React.useEffect(() => {
+    if (!poolId) {
+      setPoolSymbol("");
+      setPoolError("");
+      return;
     }
-  );
+    const couldInputBePoolId = /^(0x){1}[0-9a-f]{64}/i.test(poolId);
```
Comment (mendesfabio) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:33]: I'd rename it to `isPoolId` or `isPoolIdLike` or `isPoolIdFormat`

mendesfabio (approved):

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
Comment (luizakp) [reply packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:33]: thanks! `isPoolIdFormat` is great

devjoaov (approved):
Looking good

packages/next-app/src/utils/formatStringCase.ts
```diff
@@ -0,0 +1,19 @@
+export function toSlug(string?: string) {
```
ribeirojose, line 1: 👌

packages/next-app/src/utils/fetcher.ts
```diff
@@ -10,8 +10,16 @@ export async function fetcher<JSON = unknown>(
...
   return data;
 }
+
+export async function fetchExistingPool(poolId:string ,chainId: string) {
+  if (!poolId) return { pool: {} };
+  const data = await gql(chainId).Pool({
+    poolId: poolId,
+  });
+  return data;
+}
```
ribeirojose, line 25: still not clear why we're using this pattern instead of loading using the swr hook, but defer to you

packages/next-app/src/components/FilterDropdown.tsx
```diff
@@ -30,10 +30,10 @@ function DropdownRadioItem({
 
   function handleSelectFilter(value: string) {
     setFilterValue(value);
-    changeSelectedFilters(toSlug(filter.name), value);
+    changeSelectedFilters(toCamelCase(filter.name), value);
   }
   function handleClearFilter() {
-    clearSelectedFilter(toSlug(filter.name));
+    clearSelectedFilter(toCamelCase(filter.name));
```
ribeirojose, line 36: couldn't we simply have the filter name be camel case? I still think this pattern is very error prone

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -76,12 +84,25 @@ export function ActionAttributeContent() {
...
-                      key={field.name}
-                      label={field.name}
-                      placeholder={field.placeholder}
-                      {...register(toSlug(`${field.name}`))}
-                    />
+                    <div key={field.name}>
+                      <Input
+                        label={field.name}
+                        placeholder={field.placeholder}
+                        {...register(toCamelCase(`${field.name}`))}
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:91]: also think this is very likely to lead to errors. You're not enforcing any interfaces with TS and even so, Typescript wouldn't know if a string is or not camelCase. If you're struggling with using the same string as a name to be displayed to the end user you can just add a separate `key` property, or add a new `label` property to the selectedAction fields so that you can use it as a display name and the key as an identifier in the form.

ribeirojose (approved):
good refactors!

PR #51, opened 2023-03-28
Substitute Dropdown for Select

luizakp:
[Linear issue](https://linear.app/bleu-llc/issue/BAL-182/substitute-dropdown-for-select)

Instead of creating a dropdown with multiple filters, we are substituting it for the existing `Select` component where one value can be selected. 

Multiple selections can be implemented later on, but right now we have just one filter

https://www.loom.com/share/b7f71873e4c5487380534880482fc243


packages/next-app/src/app/daoadmin/(components)/Sidebar.tsx
```diff
@@ -82,11 +66,26 @@ export function Sidebar() {
...
+            <Select
+              onValueChange={(value: string) => setSelectedFilter(value)}
+              theme={"light"}
+            >
+              {filters.map((filter) => (
+                <SelectPrimitive.Group key={filter.name}>
+                  <SelectPrimitive.Label className="py-1 px-2 text-sm text-gray-600">
+                    {filter.name}
+                  </SelectPrimitive.Label>
+                  <SelectItem value="" theme={"light"}>
```
devjoaov, line 78: I don't have much context here but if the `filters` array has more than one element it will render another "no filter" right?

devjoaov (approved):

ribeirojose (dismissed):

mendesfabio (approved):

PR #54, opened 2023-03-29
[Pool Metadata] add "remove button" to metadata attributes table

devjoaov:
[linear issue](https://linear.app/bleu-llc/issue/BAL-183/create-remove-attribute-button)

https://www.loom.com/share/1418f112c2d94ee8901b4d090046ba05


ribeirojose (approved):
🙌  

can u leave loom videos inline in the PR comment? Makes it easier so we don't have to open another tab while reviewing

PR #56, opened 2023-03-30
add setBatchUpdateMetadata

ribeirojose:


packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -22,18 +23,48 @@ contract PoolMetadataRegistry is IPoolMetadataRegistry {
         _vault = vault;
     }
 
-    /// @notice Reverts unless the pool with the given ID is registered on Balancer.
-    /// @param poolId The ID of the pool to check registration status for.
-    modifier onlyRegisteredPool(bytes32 poolId) {
-        require(_isPoolRegistered(poolId), "Pool not registered");
+    /// @notice Wraps a single pool ID in an array.
+    /// @param poolId The pool ID to wrap.
+    function _wrapPoolId(bytes32 poolId) internal pure returns (bytes32[] memory) {
```
devjoaov, line 28: would define this function after the modifiers

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -51,18 +82,40 @@ contract PoolMetadataRegistry is IPoolMetadataRegistry {
...
+    {
+        for (uint256 i = 0; i < poolIds.length; i++) {
+            _setPoolMetadata(poolIds[i], metadataCIDs[i]);
+        }
     }
 
     /// @notice Computes the action ID for a given function selector.
     /// @param selector The function selector to compute the action ID for.
     /// @return The computed action ID.
+
```
devjoaov, line 118:
```suggestion
```

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -51,18 +82,40 @@ contract PoolMetadataRegistry is IPoolMetadataRegistry {
...
-        authenticate(poolId)
+        onlyRegisteredPools(_wrapPoolId(poolId))
+        authenticatePools(_wrapPoolId(poolId))
     {
-        poolIdMetadataCIDMap[poolId] = metadataCID;
-        emit PoolMetadataUpdated(poolId, metadataCID);
+        _setPoolMetadata(poolId, metadataCID);
+    }
+
+    function setBatchPoolMetadata(bytes32[] calldata poolIds, string[] calldata metadataCIDs)
```
devjoaov, line 104: would add documentation for this function as well
https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#tags

packages/contracts/src/PoolMetadataRegistry.sol
```diff
@@ -22,18 +23,48 @@ contract PoolMetadataRegistry is IPoolMetadataRegistry {
...
-        require(_canPerform(actionId, poolId, msg.sender), "sender not allowed");
+        for (uint256 i = 0; i < poolIds.length; i++) {
+            bytes32 poolId = poolIds[i];
+            require(_canPerform(actionId, poolId, msg.sender), "sender not allowed");
+        }
+        _;
+    }
+
+    /**
+     * @notice Validate the size of two arrays to avoid exceeding the gas limit and ensure equal length.
```
devjoaov, line 55: I think we could decide on one docstring pattern to use 
https://docs.soliditylang.org/en/v0.8.17/natspec-format.html?highlight=documentation#documentation-example

devjoaov (dismissed):
LGTM, thanks @ribeirojose. I would suggest adding a batch update test to validate that when a signer attempts to set the metadata of an array of poolIds where exists one poolId that the signer is not the owner of, the transaction would revert. Just to prevent any potential security issues.

mendesfabio (commented):
also want to see tests where the user isn't the owner of one of the pools passed in the array, otherwise looks good to me!

mendesfabio (approved):

PR #60, opened 2023-03-31
[DAO Admin] Verify gaugeAddress given by user and display symbol

luizakp:
This PR:
- refactors how the `Pool ID` validation work (now through react-hook-form) 
- added `Gauge ID` validation by adding `gauges-gql` package 

[Linear Issue](https://linear.app/bleu-llc/issue/BAL-165/verify-gaugeaddress-given-by-user-and-display-symbol)

https://www.loom.com/share/18a8f0b05426410588cf971f7e5aa99d


packages/next-app/src/utils/hardcodedData.ts
```diff
@@ -1,6 +1,57 @@
...
+
 import { ActionAttribute } from "#/contexts/AdminToolsContext";
+import gaugeGql from "#/lib/gaugesGql";
+import poolGql from "#/lib/gql";
+
+const inputField =  {
+  label: "Pool ID",
+  key: "poolId",
+  placeholder: "Insert the Pool ID here",
+  type: "string",
```
ribeirojose, line 11: I think we should remove this if this is not being used to avoid WTFs when looking at the code

packages/next-app/src/utils/hardcodedData.ts
```diff
@@ -1,6 +1,57 @@
...
+  getValidations: ( chain:
+    | (Chain & {
+        unsupported?: boolean | undefined;
+      })
+    | undefined) => ({
```
ribeirojose, line 16: equivalent to 
```suggestion
  getValidations: (chain?: Chain) => ({
```

packages/next-app/src/utils/hardcodedData.ts
```diff
@@ -1,6 +1,57 @@
...
+      const result = await poolGql(
+        chain!.id.toString()
+      ).Pool({
+        poolId: value,
+      });
+
+      return !result.pool?.symbol ? "Pool not found. Please insert an existing Pool ID" : "";
+    },
+  }),
+}
```
ribeirojose, line 28:
```suggestion
} as const
```

package.json
```diff
@@ -47,6 +46,7 @@
   },
   "homepage": "https://github.com/bleu-llc/balancer-pool-metadata#readme",
   "devDependencies": {
-    "concurrently": "^7.6.0"
+    "concurrently": "^7.6.0",
+    "ghooks": "^2.0.4"
```
ribeirojose, line 50: 😮 good catch! I think all of them should be devDependencies

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -1,5 +1,6 @@
 "use client";
 
+import _ from "lodash";
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:3]: prefer importing only the function you'll use otherwise this is likely to load the full lib to the code
```suggestion
import isEmpty from "lodash/isEmpty";
```

packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx
```diff
@@ -84,24 +115,20 @@ export function ActionAttributeContent() {
...
                         placeholder={field.placeholder}
-                        {...register(toCamelCase(`${field.name}`))}
+                        {...register(field.key, {
+                          validate: field?.getValidations?.(chain),
+                        })}
+                        errorMessage={
+                          _.isEmpty(formState.errors[field.key])
+                            ? undefined
+                            : String(formState.errors[field.key]?.message)
+                        }
```
Comment (ribeirojose) [inline packages/next-app/src/app/daoadmin/action/[actionId]/(components)/ActionAttributeContent.tsx:129]: if you wanted to do optional chaining you could also write the following
```suggestion
                        errorMessage={formState.errors?.[field.key]?.message}
```

ribeirojose (dismissed):

PR #61, opened 2023-04-03
[Pool Metadata] creates search pool button and form

devjoaov:
[linear issue](https://linear.app/bleu-llc/issue/BAL-134/select-pool-from-left-side-if-user-connected-or-search)

image

connected wallet
image

disconnected wallet
image

pool not found
image

### DIFF
--- a/packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
+++ b/packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
@@ -1,0 +1,121 @@
+"use client";
+
+import { useRouter } from "next/navigation";
+import React from "react";
+import { Controller, useForm } from "react-hook-form";
+import { useAccount, useNetwork } from "wagmi";
+
+import { Button } from "#/components/Button";
+import { Input } from "#/components/Input";
+import { Select, SelectItem } from "#/components/Select";
+import balancerGql, { networkIdEnumMap } from "#/lib/gql";
+
+export interface PoolMetadataAttribute {
+  poolId: string;
+  network: string;
+}
+
+const inputTypenames = [
+  { value: "1", label: "Mainnet" },
+  { value: "137", label: "Polygon" },
+  { value: "42161", label: "Arbitrum" },
+  { value: "5", label: "Goerli" },
+];
+
+export default function SearchPoolForm({ close }: { close?: () => void }) {
+  const {
+    register,
+    handleSubmit,
+    setError,
+    control,
+    watch,
+    clearErrors,
+    resetField,
+    formState: { errors },
+  } = useForm<PoolMetadataAttribute>();
+  const { push } = useRouter();
+  const { chain } = useNetwork();
+  const { isConnected } = useAccount();
+
+  const poolId = watch("poolId");
+  const network = isConnected ? chain?.id.toString() : watch("network");
+
+  const { data: poolAddress } = balancerGql(network || "1").usePoolAddress(
+    {
+      poolId,
+    },
+    { revalidateIfStale: true }
+  );
+
+  const isPool = poolAddress?.pool?.address ? true : false;
+
+  function handleSubmitForm(formData: PoolMetadataAttribute) {
+    const networkId = formData.network ?? chain?.id.toString();
+    const networkName =
+      networkIdEnumMap[networkId as keyof typeof networkIdEnumMap];
+    push(`/metadata/${networkName}/pool/${formData.poolId}`);
+
+    close?.();
+  }
+
+  React.useEffect(() => {
+    if (poolAddress && !poolAddress.pool && poolId) {
+      setError(
+        "poolId",
+        {
+          type: "notfound",
+          message: "Pool not found. Check the Pool ID and network.",
+        },
+        { shouldFocus: true }
+      );
+      return;
+    } else {
+      clearErrors("poolId");
+    }
+  }, [poolAddress]);
+
+  React.useLayoutEffect(() => {
+    resetField("poolId");
+  }, [network]);
+
+  return (
+    <form onSubmit={handleSubmit(handleSubmitForm)} className="px-2 pt-2">
+      {!isConnected && (
+        <>
+          <label className="mb-2 block text-sm text-gray-400">Network</label>
+          <div>
+            <Controller
+              control={control}
+              name={"network"}
+              defaultValue={"1"}
+              render={({ field: { onChange, value, ref } }) => (
+                <Select onValueChange={onChange} value={value} ref={ref}>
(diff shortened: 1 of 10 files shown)


packages/next-app/src/app/metadata/layout.tsx
```diff
@@ -19,7 +22,27 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         linkUrl={"/metadata"}
         title={"Pool Metadata"}
         imageSrc={balancerSymbol}
-      />
+      >
+        <Dialog title={"Search pool"} content={<SearchPoolForm />}>
+          <div className="pointer-events-auto relative rounded bg-white">
```
ribeirojose, line 27: Curious about why you chose to use white here. Did you want to call attention to the lighter element on a dark background?

packages/next-app/src/app/metadata/layout.tsx
```diff
@@ -19,7 +22,27 @@ export default function Layout({ children }: { children: React.ReactNode }) {
...
+              <MagnifyingGlassIcon
+                width="21"
+                height="21"
+                stroke="currentColor"
+                strokeLinecap="round"
+                strokeLinejoin="round"
+                strokeWidth={1}
+                className="mr-3 flex-none"
+              />
+              Search pool metadata...
```
ribeirojose, line 41:
```suggestion
              Search pool...
```

packages/next-app/src/components/Header.tsx
```diff
@@ -10,17 +11,19 @@ interface IHeader {
...
   return (
     <div className="flex flex-wrap items-center justify-between border-b border-gray-700 bg-gray-800 p-4 text-white">
       <Link href={linkUrl} className="mr-5 flex items-center gap-3">
         {imageSrc && <Image src={imageSrc} height={50} width={50} alt="" />}
-        <h1 className="flex gap-2 text-4xl font-thin not-italic leading-8 text-gray-200">
+        <h1 className="flex gap-2 text-xl font-thin not-italic leading-8 text-gray-200">
           Balancer <p className="font-medium">{title}</p>
         </h1>
       </Link>
+      {children && children}
```
ribeirojose, line 26: ❓

packages/next-app/src/components/Header.tsx
```diff
@@ -10,17 +11,19 @@ interface IHeader {
   linkUrl: UrlObject | __next_route_internal_types__.RouteImpl<string>;
   imageSrc?: string;
   title: string;
+  children?: ReactNode;
```
ribeirojose, line 14: Prefer to extend `React.PropsWithChildren`

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }, [poolAddress]);
+
+  React.useLayoutEffect(() => {
+    resetField("poolId");
+  }, [network]);
+
+  return (
+    <form onSubmit={handleSubmit(handleSubmitForm)} className="px-2 pt-2">
+      {!isConnected && (
```
ribeirojose, line 83: We'll be effectively making the commit make a bunch more clicks to switch network if we do this. Couldn't we always show this select?

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+        {
+          type: "notfound",
+          message: "Pool not found. Check the Pool ID and network.",
+        },
+        { shouldFocus: true }
+      );
+      return;
+    } else {
+      clearErrors("poolId");
+    }
```
ribeirojose, line 74: Any reason to set error here instead of doing it on the form submit?

ribeirojose (dismissed):
lgtm, trust your call but I don't fully grasp why add the further complication of having the user change networks when searching pools with to me is the most natural flow. Did you have any issues trying to do that?

packages/next-app/src/app/metadata/layout.tsx
devjoaov: I believe the white background would catch the user's attention, but I'm not confident with my design skills 😆

packages/next-app/src/components/Header.tsx
devjoaov: but if the header does not have children?

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
devjoaov: The button will be disabled until the user enters a valid pool id, which must correspond to an existing pool

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
devjoaov: > lgtm, trust your call but I don't fully grasp why add the further complication of having the user change networks when searching pools with to me is the most natural flow. Did you have any issues trying to do that?

@ribeirojose I faced some errors when the user chose a pool from a chain that was not the same as the selected chain. However, I am currently devising a solution to address this issue

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
ribeirojose: This is a bit unintuitive to me 😕  you're searching a pool but you can't search unless it already exists?

packages/next-app/src/components/Header.tsx
ribeirojose: No problem, this is how the type is defined:

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
devjoaov: It was a requirement we planned for the task. 

`"go to pool disabled until validation that the pool exists runs;"`

packages/next-app/src/app/metadata/layout.tsx
ribeirojose: I think it should be a different shade of the background color not to call too much attention to it

packages/next-app/src/app/metadata/(components)/SearchPoolForm.tsx
ribeirojose: Right -- this should probably be called `Go to pool` instead of search then, WDYT? Ultimately we're not showing the user a list of results, but directing them to a specific pool page

devjoaov: Agreed

PR #62, opened 2023-04-04
Create a common sidebar component

luizakp:
[Linear issue](https://linear.app/bleu-llc/issue/BAL-177/create-a-common-sidebar-component)

This PR adds:
- Sidebar generic component;
- Sidebar for Metadata and DaoAdmin
- `usePoolMetadata` as PoolMetadataContext hook

https://www.loom.com/share/80d579fd65c442bbbc45d4bf9c105f42


packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,131 @@
...
+    <div className="items-start justify-start space-y-2.5 self-stretch px-5">
+      <div className="flex items-center justify-start space-x-0 text-2xl font-medium text-gray-400">
+        <span>{name}</span>
+      </div>
+      {children}
+    </div>
+  );
+}
+
+function Content({ children }: { children: React.ReactNode }) {
```
ribeirojose, line 66: Prefer `React.PropsWithChildren`

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function Header({
+  children,
+  name,
+}: {
+  children?: React.ReactNode;
+  name: string;
+}) {
+  useSidebarContext();
```
ribeirojose, line 55: ❓

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function InputFilter({
+  placeHolder,
+  onChange,
+}: {
+  placeHolder: string;
+  onChange: (e: React.FormEvent<HTMLInputElement>) => void;
+}) {
+  useSidebarContext();
```
ribeirojose, line 87: 👀 is this supposed to be here?

packages/next-app/src/app/metadata/(components)/MetadataSidebar.tsx
```diff
@@ -0,0 +1,74 @@
...
+              onClick={() => handleSetPool(item.id)}
+            >
+              <Sidebar.Item isSelected={item.id === selectedPool}>
+                <OwnedPool
+                  isSelected={item.id === selectedPool}
+                  pool={item as Pool}
+                />
+              </Sidebar.Item>
+            </Link>
+          ))}
```
ribeirojose, line 34: Could we add a loading state here?

packages/next-app/src/app/daoadmin/(components)/DaoAdminSidebar.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+  ];
+
+  return (
+    <Sidebar>
+      <Sidebar.Header name="DAO Actions">
+        <div className="flex justify-between">
+          <Sidebar.InputFilter
+            placeHolder="Search for DAO Action"
+            onChange={handleSearchChange}
```
ribeirojose, line 51:
```suggestion
            onChange={(e) => setQuerySearch(e.currentTarget.value)}
```

packages/next-app/src/app/daoadmin/(components)/DaoAdminSidebar.tsx
```diff
@@ -0,0 +1,103 @@
...
+          />
+          <Select
+            onValueChange={(value: string) => setSelectedFilter(value)}
+            theme={"light"}
```
ribeirojose, line 55:
```suggestion
            theme="light"
```

packages/next-app/src/app/daoadmin/(components)/DaoAdminSidebar.tsx
```diff
@@ -0,0 +1,103 @@
...
+                </SelectPrimitive.Label>
+                <SelectItem value="" theme={"light"}>
+                  No filter
+                </SelectItem>
+                {filter.options.map((option) => (
+                  <SelectItem key={option} value={option} theme={"light"}>
+                    {option}
+                  </SelectItem>
+                ))}
+              </SelectPrimitive.Group>
```
ribeirojose, line 70: Wouldn't it make sense to expose this through the Select component instead of having both here?

ribeirojose (approved):

packages/next-app/src/components/Sidebar.tsx
luizakp: I thought it was overcomplicating doing a new component just for the Header Title which has this code: 

```
<div className="flex items-center justify-start space-x-0 text-2xl font-medium text-gray-400">
    <span>{name}</span>
</div>
```

And I used the name as a props. But now thinking about it the user might not want to have a Header title, and with this I'm forcing it...

My new approach would be:

```
function Header({
  children
}: {
React.PropsWithChildren) {
  useSidebarContext();
  return (
    <div className="items-start justify-start space-y-2.5 self-stretch px-5">
      {children}
    </div>
  );
}

function HeaderTitle({
children
}:{React.PropsWithChildren}) {
useSidebarContext();
return(
  <div className="flex items-center justify-start space-x-0 text-2xl font-medium text-gray-400">
    <span>{name}</span>
  </div>
)}
```

WDYT?

packages/next-app/src/components/Sidebar.tsx
luizakp: Do you mean the `useSidebarContext();` or the `InputFilter`?  If we see `InputFilter` being applicable somewhere else then I would transfer this as a standalone component

packages/next-app/src/app/metadata/(components)/MetadataSidebar.tsx
luizakp: We currently have a loading state from the main fetch that gathers the pool data. We could pass here using context and using the loading state inside the `Sidebar` but right now I don't think it would be useful as the page is only rendered if the `poolsData` length is bigger than 0. With the current way we are doing the fetching we only update `poolsData` state when the promise is over. So there is no way the Sidebar will be rendered without the complete data

image

I agree we should have a loading state here, but for that I think we should first check if it's possible to gql to kind of parse the response into chunks and update those chunks into the `poolsData` state.

Does that make sense to you?

packages/next-app/src/components/Sidebar.tsx
ribeirojose: I mean what's the purpose of calling this hook here with no return? 🥸

packages/next-app/src/components/Sidebar.tsx
ribeirojose: I think that makes sense -- I also like this approach better than the first one!

packages/next-app/src/components/Sidebar.tsx
luizakp: aaa this hook is to make sure that sidebar sub-components cannot be used outside `<Sidebar>`

without it, you would be able to use these components everywhere without `<Sidebar>`
Sidebar.Header = Header;
Sidebar.Content = Content;
Sidebar.InputFilter = InputFilter;
Sidebar.Item = Item;

ribeirojose: And would that really be an issue in your opinion? 😅

PR #71, opened 2023-04-13
Improve write contract hooks

luizakp:
[Linear issue](https://linear.app/bleu-llc/issue/BAL-285/improve-write-contract-hooks)


packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +153,100 @@ export default function Page() {
...
+  token: {
+    __typename?: "UserInternalBalance" | undefined;
+    token: `0x${string}`;
+    balance: string;
+  };
+  userAddress: `0x${string}`;
+  setTransaction: React.Dispatch<React.SetStateAction<ITransaction>>;
+  chain?: Chain;
+}) {
+  const [transferKind, setTransferKind] = useState<UserBalanceOpKind>();
```
mendesfabio, line 172: this sounds confusing to me since transfer is one of the operations - what about `operationKind` instead?

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +153,100 @@ export default function Page() {
...
+  function handleWithdraw() {
+    setTransferKind(UserBalanceOpKind.WITHDRAW_INTERNAL);
+    setTransaction((prev) => ({
+      ...prev,
+      status: TransactionStatus.WAITING_APPROVAL,
+    }));
+  }
+
+  useEffect(() => {
+    if (!data) return;
```
mendesfabio, line 194: could add `tiny-invariant` for these checks

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -1,15 +1,26 @@
...
 import { Toast } from "#/components/Toast";
 import { impersonateWhetherDAO, pools } from "#/lib/gql";
 import { tokenDictionary } from "#/utils/getTokenInfo";
-import { writeWithdrawInternalBalance } from "#/wagmi/withdrawInternalBalance";
+import {
+  usePrepareVaultManageUserBalance,
+  useVaultManageUserBalance,
+} from "#/wagmi/generated";
+
+export enum UserBalanceOpKind {
```
mendesfabio, line 18: would probably consolidate these constants and any helper functions at `lib` or `utils`

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +153,100 @@ export default function Page() {
...
+    kind: transferKind as number,
+    asset: token.token,
+    amount: parseFixed(token.balance, tokenDictionary[token.token].decimals),
+    sender: userAddress as `0x${string}`,
+    recipient: userAddress as `0x${string}`,
+  };
+  const { config } = usePrepareVaultManageUserBalance({
+    args: [[userBalanceOp]],
+  });
+  const { data, write } = useVaultManageUserBalance(config);
```
mendesfabio, line 183: 😎

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +153,100 @@ export default function Page() {
...
+
+  function handleWithdraw() {
+    setTransferKind(UserBalanceOpKind.WITHDRAW_INTERNAL);
+    setTransaction((prev) => ({
+      ...prev,
+      status: TransactionStatus.WAITING_APPROVAL,
```
mendesfabio, line 189: wondering if we shouldn't use [useWaitForTransaction](https://wagmi.sh/react/hooks/useWaitForTransaction) to help us tracking the status of a submitted transaction

mendesfabio (dismissed):

packages/next-app/src/app/internalmanager/page.tsx
luizakp: hmmmm not quite sure if `tiny-invariant` do what we want here. I just want to check if data exists before trying to manipulate data. But if it doesn't exist I just want to skip the rest of the function for now as this is an useEffect.  If I use `tiny-invariant` it'll throw the error on the first rendering as data is still undefined

luizakp:
@mendesfabio still thinking if using `useWaitForTransaction`  for the management of the transaction status 
impairs the understanding of the flow or not because it's disconnected from the rest 🤔

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +147,100 @@ export default function Page() {
...
+    }));
+  }
+
+  useEffect(() => {
+    if (!data) return;
+    const { hash } = data;
+    switch (operationKind) {
+      case UserBalanceOpKind.WITHDRAW_INTERNAL: {
+        async function handleTransactionStatus() {
```
ribeirojose, line 208: Did you mean to define this function here?

packages/next-app/src/app/internalmanager/page.tsx
```diff
@@ -188,3 +147,100 @@ export default function Page() {
...
+          if (hash && chain) {
```
ribeirojose, line 209: Could return early empty when one of the two is false

ribeirojose (dismissed):

packages/next-app/src/app/internalmanager/page.tsx
mendesfabio: question is what would cause the `data` to not exist then - is the wagmi generated hook a promise? trying to generate it locally but getting an error

luizakp: By default, the `useContractWrite()` from wagmi core has data as undefined unless a transaction is occurring
image

mendesfabio: hmm ok, thought we should revert if we don't have `data` but it behaves differently 🤔 still thinking about this useEffect though, wagmi should handle it natively but if we don't find a better solution that's fine too.

PR #73, opened 2023-04-20
[Internal Manager] Connection to mainnet

luizakp:
This PR adds mainnet support to internal manager app

### DIFF
--- a/packages/gql/codegen.ts
+++ b/packages/gql/codegen.ts
@@ -70,7 +70,7 @@ export const SUBGRAPHS = {
       return {
         //TODO: deploy subgraph on mainnet, polygon and arbitrum
         // https://linear.app/bleu-llc/issue/BAL-290/deploy-subgraph-with-token-relation-on-other-networks
-        [Network.Mainnet]: `${baseEndpoint}/balancer-v2-goerli`,
+        [Network.Mainnet]: `https://api.thegraph.com/subgraphs/name/bleu-llc/balancer-goerli-v2`,
         [Network.Polygon]: `${baseEndpoint}/balancer-v2-goerli`,
         [Network.Arbitrum]: `${baseEndpoint}/balancer-v2-goerli`,
         [Network.Goerli]: `${baseEndpoint}/balancer-v2-goerli`,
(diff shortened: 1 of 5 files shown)


luizakp:
This is a quick PR just to attend @mendesfabio needs. We `#must` re-deploy the subgraph on the mainnet with the correct name and have a function to re-fetch the data from gql when the chain changes. Right now you have to reload the page manually

packages/gql/codegen.ts
```diff
@@ -70,7 +70,7 @@ export const SUBGRAPHS = {
       return {
         //TODO: deploy subgraph on mainnet, polygon and arbitrum
         // https://linear.app/bleu-llc/issue/BAL-290/deploy-subgraph-with-token-relation-on-other-networks
-        [Network.Mainnet]: `${baseEndpoint}/balancer-v2-goerli`,
+        [Network.Mainnet]: `https://api.thegraph.com/subgraphs/name/bleu-llc/balancer-goerli-v2`,
```
ribeirojose, line 73: Is this really supposed to be Goerli?

packages/next-app/src/wagmi/client.ts
```diff
@@ -1,24 +1,31 @@
...
 
-const { chains, provider, webSocketProvider } = configureChains(
-  [polygon, goerli],
-  [publicProvider()]
-);
+export function getWagmiClient(routerPath: string) {
+  const walletChains =
+    routerPath === "/internalmanager"
+      ? [mainnet, goerli, polygon]
+      : [polygon, goerli];
```
ribeirojose, line 11: 🤔 we could leave mainnet in all paths, but would just need show a message saying it's unsupported yet.

packages/next-app/src/components/RootLayout.tsx
```diff
@@ -2,14 +2,17 @@
...
 
 import { WagmiConfig } from "#/wagmi";
-import { chains, client } from "#/wagmi/client";
+import { getWagmiClient } from "#/wagmi/client";
 
 import { ToastProvider } from "./Toast";
 
 export function RootLayout({ children }: React.PropsWithChildren) {
+  const path = usePathname();
+  const { client, chains } = getWagmiClient(path);
```
ribeirojose, line 15: 👍 got it. A comment explaining the rationale behind using this would've been helpful here

ribeirojose (approved):

packages/gql/codegen.ts
luizakp: it's with the wrong name but it's deployed on the mainnet xD we have to redeploy with the correct name later, or that can be done automatically when Balancer accepts Fabio's changes on their repo

PR #77, opened 2023-04-25
[Internal Manager] Adapt to accept transfer workflow

luizakp:
This PR adds UI and form validator support for transfers

https://www.loom.com/share/e6b3c1d3eb7f4eb68684c0d53623f8a1


packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
```diff
@@ -76,28 +82,32 @@ export default function Page({
     switch (operationKind) {
```
Comment (mendesfabio) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:82]: why did you create this switch case? any benefits you see? I personally prefer to do this using enum + dict/record

mendesfabio (dismissed):

PR #79, opened 2023-04-26
add deploy-preview workflow

ribeirojose:


.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,39 @@
+name: Deploy PR preview app to Fly
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, closed]
+
```
mendesfabio, line 6: think we should filter by path, don't need to deploy the app on every PR

```
    paths:
      - "packages/next-app/**"
 ```

mendesfabio: same applies to other actions actually, we're running foundry on PRs that only changes the Next app

mendesfabio (approved):

.github/workflows/deploy-preview.yml
ribeirojose: We're seldom making PRs w/o changes to the next app, so I think that's not as relevant right now. Would probably spend more time than save so I'll add both suggestions as improvements to our backlog

PR #81, opened 2023-04-27
[Balancer Tools] add UI to unsupported chains in certain

luizakp:
https://www.loom.com/share/5b7233b2885341c6a2e10f660420fc6c


mendesfabio (commented):
looks good, but I would take the opposite approach: instead of passing the `unsupported` chain would pass the `supported` - also think this should be a list instead of a single item. please, correct me if I'm wrong, but I'm just afraid we add X chain to the network switch and forget to config that Internal Manager doesn't support, breaking the app there. What do you think @luizakp?

luizakp:
> looks good, but I would take the opposite approach: instead of passing the `unsupported` chain would pass the `supported` - also think this should be a list instead of a single item. please, correct me if I'm wrong, but I'm just afraid we add X chain to the network switch and forget to config that Internal Manager doesn't support, breaking the app there. What do you think @luizakp?

Agree it should be a list. About being `supported` we will have to remember to add too, if we forget instead of breaking the app, the user will think the app is not available on the X chain, which I think is better

packages/next-app/src/app/internalmanager/layout.tsx
```diff
@@ -20,12 +21,12 @@ export default function Layout({ children }: { children: React.ReactNode }) {
           imageSrc={balancerSymbol}
         />
         <InternalManagerProvider>
-          <CheckUnsupportedChain
-            unsupportedChain="Polygon"
-            chainName={chain?.name}
+          <CheckSupportedChains
+            supportedChains={["Ethereum", "Goerli"]}
```
mendesfabio, line 25: I remember we have a `Network` enum somewhere, should instead use it and make any changes if needed because e.g. here we have Ethereum and there IIRC Mainnet.

mendesfabio (approved):
Just one last improvement and then feel free to merge =)

PR #82, opened 2023-04-29
[Balancer Tools] refetch gql request when chain or user changes

luizakp:
https://www.loom.com/share/d263b587408944b3b09b88eff4a6453e

### DIFF
--- a/packages/gql/src/balancer-internal-manager/internalBalances.ts
+++ b/packages/gql/src/balancer-internal-manager/internalBalances.ts
@@ -9,7 +9,6 @@ export const internalBalance = gql`
           name
           symbol
           address
-          decimals
         }
         balance
       }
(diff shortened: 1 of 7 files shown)


packages/gql/src/balancer-internal-manager/internalBalances.ts
```diff
@@ -9,7 +9,6 @@ export const internalBalance = gql`
           name
           symbol
           address
-          decimals
```
mendesfabio, line 12: not sure if I understood why you removed the decimals here

mendesfabio (approved):

packages/gql/src/balancer-internal-manager/internalBalances.ts
mendesfabio: nvm! you don't need the decimals here, right? this is the query you use to display user's internal balances and `balance` is already downscaled so you don't have to parse it.

PR #91, opened 2023-05-11
[Balancer Tools] Nested route according to chain

luizakp:
This PR adds:
- Url according to chain;
- Enum Network.Mainnet to Network.Ethereum (makes integration with wagmi easier);
- Update app url when chain changes;

https://www.loom.com/share/3f4ffe2a222340baa83da74e44a860fc


packages/next-app/src/app/internalmanager/layout.tsx
```diff
@@ -1,28 +1,40 @@
...
         <InternalManagerProvider>
           <CheckSupportedChains
-            supportedChains={["Ethereum", "Goerli", "Sepolia"]}
+            supportedChains={[
+              "Ethereum",
+              "Gnosis",
+              "Arbitrum One",
+              "Optimism",
+              "Goerli",
+              "Sepolia",
```
mendesfabio, line 36: would refactor this to use the `Network` enum

packages/next-app/src/app/internalmanager/layout.tsx
```diff
@@ -1,28 +1,40 @@
...
+
+  const network = getNetwork(chain?.name);
+
   return (
     <NetworksContextProvider>
       <div className="flex h-full flex-col">
+        <HeaderNetworkMismatchAlert />
         <Header
-          linkUrl={"/internalmanager"}
+          linkUrl={`/internalmanager/${network}` as Route}
```
ribeirojose, line 24: if eslint is not recognizing this as a route there's probably something weird with the `network` type 🤔

packages/next-app/src/contexts/networks.tsx
```diff
@@ -1,9 +1,21 @@
...
 
 import { useNetwork } from "#/wagmi";
 
+export function getNetwork(chainName?: string) {
+  if (!chainName) return Network.Ethereum;
+  const network =
+    chainName?.toLowerCase() === "arbitrum one"
+      ? "arbitrum"
+      : chainName?.toLowerCase();
+  return network as Network;
```
ribeirojose, line 16: 👁️

packages/gql/codegen.ts
```diff
@@ -15,11 +16,13 @@ export const SUBGRAPHS = {
       const baseEndpoint = "https://api.thegraph.com/subgraphs/name/bleu-llc";
 
       return {
-        [Network.Mainnet]: `${baseEndpoint}/balancer-pools-metadata`,
+        [Network.Ethereum]: `${baseEndpoint}/balancer-pools-metadata`,
```
ribeirojose, line 19: Can you leave a comment on the top of this file explaining why we chose to use this pattern here? I think it'd be very hard for a newcomer to reason about this "gambi" 😅

ribeirojose (approved):

packages/gql/codegen.ts
luizakp: It's not clear to me which pattern I should do a comment xD sorry

ribeirojose: Sorry, should've been more explicit here but I think you could also have thought of what would generate [WTFs](https://miro.medium.com/v2/resize:fit:1000/0*HaqsruEqnR5pxQD4) in this file. We're pointing a bunch networks to a Goerli endpoint -- I assume because you didn't want to deal with TS errors right now -- that's a non obvious choice that I think must be documented here.

packages/next-app/src/app/internalmanager/layout.tsx
ribeirojose: probably related to nextjs' typedRoutes -- you can disable that in nextjs.config.js if it's getting in the way

packages/next-app/src/contexts/networks.tsx
```diff
@@ -26,11 +38,21 @@ export const NetworksContextProvider = ({
...
-
     setSelectedNetwork(network.chain?.id);
+    if (
+      network.chain?.id &&
+      networkConnectedToWallet !== undefined &&
+      networkConnectedToWallet !== network.chain?.id
+    ) {
+      push(
+        `/${appName}/${networkFor(network.chain.id).toLowerCase()}` as Route
+      );
```
ribeirojose, line 54: do you think we could also preserve the rest of the URL (after the chain path)?

ribeirojose (dismissed):

packages/gql/codegen.ts
luizakp: AAA okkk on it

mendesfabio (dismissed):

mendesfabio (dismissed):

mendesfabio (approved):

PR #92, opened 2023-05-12
Pedro/bal 342 refactor ampfactor stableswap math to

yvesfracari:
This PR closes BAL-337, BAL-338, BAL-339, BAL-340, BAL-341, and BAL-342


packages/tsconfig/base.json
```diff
@@ -3,6 +3,7 @@
   "display": "Default",
   "compilerOptions": {
     "rootDir": "../../",
+    "target": "es2020",
```
ribeirojose, line 6: 👁️

packages/math/metaStable.ts
```diff
@@ -0,0 +1,45 @@
+import { OldBigNumber } from "@balancer-labs/sor";
```
ribeirojose, line 1: 🤔 I'm a bit out of context here but how many BN implementations are we currently using?

packages/math/stable.ts
```diff
@@ -0,0 +1,172 @@
...
+  const D = StableMaths._invariant(A, balances);
+  let S = ZERO;
+  for (let i = 0; i < totalCoins; i++) {
+    if (i != tokenIndexIn && i != tokenIndexOut) {
+      S = S.plus(balances[i]);
+    }
+  }
+  const x = balances[tokenIndexIn];
+  const y = balances[tokenIndexOut];
+  // A is passed as an ethers bignumber while maths uses bignumber.js
```
ribeirojose, line 34: Couldn't we require this to happen before an expect a bignumber.js object here?

yvesfracari: This function was copied from @balancer/sor package, since was not exported

packages/math/metaStable.ts
yvesfracari: The @balancer/sor package uses a self-defined version (OldBigNumber) and one from ethersproject (BigNumber). I thought would be good to follow the same pattern here

ribeirojose (approved):

PR #94, opened 2023-05-15
[Internal Manager] Handle user inputted token address and add OG Image

luizakp:
https://linear.app/bleu-llc/issue/BAL-324/handle-user-inputted-token-address

https://linear.app/bleu-llc/issue/BAL-365/add-og-image-and-handle-ui-direct-link-to-token

https://www.loom.com/share/f9bf4e2ad5dc4c8095265a63c3baf7c3


packages/shared/index.ts
```diff
@@ -42,6 +42,17 @@ export function buildBlockExplorerTxURL({
   return `${networkUrl.url}tx/${txHash}`;
 }
 
+export function buildExplorerTokenURL({
```
mendesfabio, line 45: to keep the convention
```suggestion
export function buildBlockExplorerTokenURL({
```

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
```diff
@@ -39,16 +40,22 @@ import {
...
+>;
+
 export default function Page({
   params,
 }: {
   params: {
     tokenAddress: `0x${string}`;
     network: Network;
-    operationKind: operationKindType;
+    operationKind: "deposit" | "withdraw" | "transfer";
```
Comment (mendesfabio) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:54]: doesn't `keyof typeof operationKindType` work here?

mendesfabio (approved):

packages/shared/index.ts
```diff
@@ -42,6 +42,17 @@ export function buildBlockExplorerTxURL({
...
+  chainId,
+  tokenAddress,
+}: {
+  chainId: NetworkChainId;
+  tokenAddress: `0x${string}`;
```
ribeirojose, line 50: `#could` create a singular type to use across all `0x{string}` usages

packages/next-app/src/app/ogImage.tsx
```diff
@@ -0,0 +1,104 @@
+export function OgImage({ appName }: { appName: string }) {
```
ribeirojose, line 1: Would move this to components

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
```diff
@@ -85,6 +88,22 @@ export default function Page({
...
+    const tokenInfo = internalBalanceTokenData?.user
+      ?.userInternalBalances?.[0] || {
+      balance: "0",
+      tokenInfo: {
+        __typename: "Token",
+        address: params.tokenAddress,
+        decimals: walletAmount?.decimals || 0,
+        symbol: walletAmount?.symbol || "",
+        name: walletAmount?.symbol || "",
+      },
```
Comment (ribeirojose) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:101]: could you say more to why this is needed here? Should we even be rendering this component if the user doesn't have any token balances?

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
```diff
@@ -85,6 +88,22 @@ export default function Page({
...
+    };
+
+    setTokenData(tokenInfo);
+  }, []);
```
Comment (ribeirojose) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:105]: shouldn't `internalBalanceTokenData` be in your dependencies array?

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
```diff
@@ -39,16 +40,22 @@ import {
 import { refetchRequest } from "#/utils/fetcher";
 import { ArrElement, GetDeepProp } from "#/utils/getTypes";
 
+type TokenData = Omit<
+  ArrElement<GetDeepProp<SingleInternalBalanceQuery, "userInternalBalances">>,
+  "__typename"
+>;
```
Comment (ribeirojose) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:46]: `__typename` shouldn't be added to responses unless we're explicitly querying for it. Is this not what you've observed?

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
Comment (ribeirojose) [inline packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:101]: It sounds like this would have the same effect of assigning to the variable outside of the useEffect if you're not setting tokenData elsewhere

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -212,7 +314,9 @@ function TokenRow({
     <Table.BodyRow key={token.tokenAddress}>
       <Table.BodyCell customWidth="w-12">
         <Link
-          href={`/internalmanager/${chainName}/${operationKind}/${token.tokenAddress}`}
+          href={
+            `/internalmanager/${chainName}/${operationKind}/${token.tokenAddress}` as Route
```
ribeirojose, line 318: there's something off if this isn't been recognized as a route -- lowpriority #techdebt ticket imo

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+      return false;
+    }
+  }
+
   return (
     <div className="text-white divide-y divide-gray-700 max-h-[30rem] overflow-y-scroll scrollbar-thin scrollbar-thumb-gray-200 scrollbar-track-blue3">
       <div className="w-full flex flex-col justify-center items-center h-full py-4 gap-y-4">
         <div className="text-xl">Token Search</div>
+        <div className="w-full px-10 flex items-center">
+          <input
```
ribeirojose, line 232: Any reason not to use the standard input we have?

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -80,6 +85,9 @@ function TokenModal({
   let { address } = useAccount();
   address = impersonateWhetherDAO(chain?.id.toString() || "1", address);
   const [tokens, setTokens] = useState<(TokenItem | undefined)[]>([]);
+  const [tokenSearch, setTokenSearch] = useState<string>("");
```
ribeirojose, line 88: no need to say it's a string if you're initializing its value

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -80,6 +85,9 @@ function TokenModal({
...
+  const [isTokenSearchDisabled, setIsTokenSearchDisabled] =
```
ribeirojose, line 89: same as above

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
     }
   }
 
+  const tokenExplorerUrl = buildExplorerTokenURL({
+    chainId: chain!.id,
```
ribeirojose, line 176: #could make this not required and throw an error or return empty if `chain?.id` is not present
```suggestion
    chainId: chain.id,
```

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+    tokenAddress: tokenSearch.toLowerCase() as `0x${string}`,
```
ribeirojose, line 177: `#should` already be setting it o lowercase on the input change

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+  });
+
   useEffect(() => {
     if (!internalBalanceData?.user?.userInternalBalances) return;
     getWalletBalance(tokenAdresses as `0x${string}`[]);
   }, [internalBalanceData]);
 
+  useEffect(() => {
+    if (addressRegex.test(tokenSearch)) {
```
ribeirojose, line 186: `#would` (style) return early if the condition is not met to avoid indenting multiple conditions. also that follows the same pattern you used in lines 218-223

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+      setIsTokenSearchDisabled(false);
+      if (
+        tokens.filter((token) => filterTokenInput({ tokenSearch, token }))
+          .length < 1
```
ribeirojose, line 190:
```suggestion
        tokens.some((token) => filterTokenInput({ tokenSearch, token }))
```

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+      ) {
+        fetchSingleTokenBalance({
+          tokenAddress: tokenSearch.toLowerCase() as `0x${string}`,
+        }).then((tokenData) => {
+          setTokens((prev) => [
+            ...prev,
+            {
+              ...tokenData,
+              name: tokenData.symbol,
+              tokenAddress: tokenSearch.toLowerCase() as `0x${string}`,
```
ribeirojose, line 200: not sure I follow -- wouldn't these be have been populated already? Also any reason why we're using `name` instead of the more standard `symbol` for displaying the token nicely on the UI?

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+  }: {
+    tokenSearch: string;
+    token?: TokenItem;
+  }) {
+    {
+      if (tokenSearch === "") return true;
+      const searchString = tokenSearch.toLowerCase();
+      if (token?.name?.toLowerCase().includes(searchString)) return true;
+      if (token?.symbol.toLowerCase().includes(searchString)) return true;
+      if (token?.tokenAddress.toLowerCase().includes(searchString)) return true;
```
ribeirojose, line 222: `#comment` less conventional, but I'd do something in the lines of:

```suggestion
      if (!token) return false;
      const regex = new RegExp(tokenSearch, "i"); // match tokenSearch, case insensitive
      return regex.test(Object.values(token).join(","))
```

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+
+  function filterTokenInput({
+    tokenSearch,
+    token,
+  }: {
+    tokenSearch: string;
+    token?: TokenItem;
+  }) {
+    {
+      if (tokenSearch === "") return true;
```
ribeirojose, line 218: why would you return true if tokenSearch is ? This smells like something that shouldn't be here

ribeirojose (dismissed):
left a few remarks

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
Comment (luizakp) [reply packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:101]: This is the case where the user doesn't have the token as an Internal Balance but has the token in their wallet, so the user can still do a deposit even with no internal balance (which is `balance`)

packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx
Comment (luizakp) [reply packages/next-app/src/app/internalmanager/[network]/[operationKind]/[tokenAddress]/page.tsx:46]: Actually no, the __typename is always there on the responses and is also expected from the type created by gql. 

In case of only having the token on the wallet we are creating the TokenData from wagmi info instead of gql and that's why there is no `__typename`

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
luizakp: They have different designs, we have to refactor the Input to accept variants 

https://linear.app/bleu-llc/issue/BAL-381/input-style-variants

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -156,15 +172,100 @@ function TokenModal({
...
+            type="text"
+            placeholder="Search name or paste address"
+            className="bg-blue4 h-9 w-full appearance-none items-center justify-center rounded-l-[4px] px-[10px] text-sm leading-none text-slate12 outline-none"
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              setTokenSearch(e.target.value)
+            }
+            value={tokenSearch}
+          />
+          {isTokenSearchDisabled ? (
+            <button
```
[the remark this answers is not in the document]
luizakp: the proposed design has a style that is different from the base style of the standard button. We have to refactor the standard button if you think that makes sense

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
luizakp: this is the case for an inputted token that doesn't show on the token list and the user only has this token on their wallet. Therefore, we don't have this data already

`name` is not provided from the wagmi `fetchBalance` so we are using `symbol ` to not do an extra fetch jsut for the name. But the tokens from the InternalBalance and token list have their names showing

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -112,13 +120,21 @@ function TokenModal({
...
+    tokenAddress,
+  }: {
+    tokenAddress: `0x${string}`;
+  }) {
+    const tokenData = await fetchBalance({
+      address: addressLower as `0x${string}`,
+      token: tokenAddress,
+    });
+    return tokenData;
+  }
```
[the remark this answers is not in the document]
luizakp: I'm using thins function nside of an useEffect as well, so I created this to be more clear and use on both occasions (inside and outside of the useEffect)

mendesfabio (approved):

PR #98, opened 2023-05-18
Pedro/bal 384 import pool modal

yvesfracari:
This PR:
- Refactor GQL queries (fewer queries and more generics)
- Refactor Search Pool Form (make it easy to use in other contexts)
- Add the import pool modal with the function to load the pool variables
- Impacts the metadata application as well

https://www.loom.com/share/41d16418dc8c4bd7adead14ebc6d3d04


packages/gql/src/balancer-pools/pools.ts
```diff
@@ -1,7 +1,7 @@
 import { gql } from "graphql-tag";
 
 export const poolWhereOwner = gql`
-  query Pools($owner: Bytes!) {
+  query PoolsWhereOwner($owner: Bytes!) {
```
ribeirojose, line 4: `#would` call this only `Pools`

packages/gql/src/balancer-pools/pools.ts
```diff
@@ -15,26 +15,43 @@ export const poolWhereOwner = gql`
...
+    $poolId: ID!
+    $poolTypes: [String!] = [
+      "Weighted"
+      "ComposableStable"
+      "Stable"
+      "MetaStable"
+      "Element"
+      "LiquidityBootstrapping"
+      "Linear"
+    ]
```
ribeirojose, line 29: 👀 `TIL`

packages/next-app/src/app/metadata/layout.tsx
```diff
@@ -33,7 +36,12 @@ export default function Layout({ children }: React.PropsWithChildren) {
         <div className="flex flex-1 gap-x-8">
           <div>
             <Sidebar isFloating>
-              <Dialog title="Go  to pool" content={<SearchPoolForm />}>
+              <Dialog
+                title="Go to pool"
+                content={
+                  <SearchPoolForm onSubmit={usePoolMetadata().handleGoToPool} />
```
ribeirojose, line 42: 👀

packages/next-app/src/contexts/PoolMetadataContext.tsx
```diff
@@ -88,6 +93,16 @@ export function PoolMetadataProvider({ children }: PropsWithChildren) {
...
+  function handleGoToPool(formData: PoolAttribute) {
+    const { push } = useRouter();
+    const { chain } = useNetwork();
+
+    const networkId = formData.network ?? chain?.id.toString();
+    const networkName =
+      networkIdEnumMap[networkId as keyof typeof networkIdEnumMap];
+    push(`/metadata/${networkName}/pool/${formData.poolId}`);
+  }
+
```
ribeirojose, line 105: Not sure I follow why define this function within the context -- are we using anything that's context dependent here?

packages/next-app/src/components/WalletNotConnected.tsx
```diff
@@ -19,7 +20,12 @@ export default function WalletNotConnected({
         <div className="text-center text-slate11 text-lg">
           {/* TODO: */}
           Alternatively, open a pool directly clicking&nbsp;
-          <Dialog title="Go to pool" content={<SearchPoolForm />}>
+          <Dialog
+            title="Go to pool"
+            content={
+              <SearchPoolForm onSubmit={usePoolMetadata().handleGoToPool} />
```
ribeirojose, line 26: 👀

packages/next-app/src/components/SearchPoolForm.tsx
```diff
@@ -33,33 +38,27 @@ export default function SearchPoolForm({ close }: { close?: () => void }) {
...
 
-  function handleSubmitForm(formData: PoolMetadataAttribute) {
-    const networkId = formData.network ?? chain?.id.toString();
-    const networkName =
-      networkIdEnumMap[networkId as keyof typeof networkIdEnumMap];
-    push(`/metadata/${networkName}/pool/${formData.poolId}`);
+  const isPool = !!poolData?.pools?.length;
 
+  function handleSubmitForm(formData: PoolAttribute) {
+    onSubmit?.(formData);
```
ribeirojose, line 56: 👍 Great addition -- I'd add that if `onSubmit` is not defined this should probably throw an error

packages/next-app/src/components/SearchPoolForm.tsx
```diff
@@ -33,33 +38,27 @@ export default function SearchPoolForm({ close }: { close?: () => void }) {
...
-      poolId,
-    },
-    { revalidateIfStale: true }
-  );
-
-  const isPool = poolAddress?.pool?.address ? true : false;
+  const gqlVariables = poolTypeFilter?.length
+    ? { poolId, poolTypes: poolTypeFilter }
+    : { poolId };
+  const { data: poolData } = pools
```
ribeirojose, line 49: this is probably `poolsData` right?

packages/next-app/src/components/SearchPoolForm.tsx
```diff
@@ -33,33 +38,27 @@ export default function SearchPoolForm({ close }: { close?: () => void }) {
...
+    .gql(network || "1")
+    .usePoolsWherePoolTypeInAndId(gqlVariables, { revalidateIfStale: true });
```
ribeirojose, line 51: Do we really need to revalidate? 🤔 when would this get stale?

packages/next-app/src/components/SearchPoolForm.tsx
```diff
@@ -33,33 +38,27 @@ export default function SearchPoolForm({ close }: { close?: () => void }) {
...
-    {
-      poolId,
-    },
-    { revalidateIfStale: true }
-  );
-
-  const isPool = poolAddress?.pool?.address ? true : false;
+  const gqlVariables = poolTypeFilter?.length
+    ? { poolId, poolTypes: poolTypeFilter }
+    : { poolId };
```
ribeirojose, line 48: 👍 Could also be 
```suggestion
  const gqlVariables = { poolId, ...(poolTypeFilter.length ? {poolTypes: poolTypeFilter} : {}) }
```

packages/gql/src/balancer-pools/pools.ts
yvesfracari: Merge this with which query?

packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+          rate: token?.priceRate,
+        })) || [],
+    };
+  }
+
+  async function handleImportPoolParametersById(formData: PoolAttribute) {
+    const poolData = await pools.gql(formData.network || "1").Pool({
+      poolId: formData.poolId,
+    });
+    setInitialData(convertGqlToAnalysisData(poolData));
```
ribeirojose, line 50: to make convertGqlToAnalysisData less complicated you can also do
```suggestion
    if (!poolData) return
    setInitialData(convertGqlToAnalysisData(poolData));
```

packages/next-app/src/components/Header.tsx
```diff
@@ -17,9 +17,10 @@ interface IHeader {
   imageSrc?: string;
   title: string;
   children?: ReactNode;
+  wallet?: boolean;
 }
 
-export function Header({ linkUrl, imageSrc, title, children }: IHeader) {
+export function Header({ linkUrl, imageSrc, title, children, wallet=true }: IHeader) {
```
ribeirojose, line 23: prettier didn't run here 😅

packages/next-app/src/contexts/PoolMetadataContext.tsx
yvesfracari: It is not, but it was a way to share the function between the layout (sidebar) and the pages. I could also put it inside a handler folder. WDYT?

ribeirojose (dismissed):

packages/next-app/src/contexts/PoolMetadataContext.tsx
ribeirojose: I'd suggest only keep things in the context that need to be there and if want to share that in two files just separate it into another file

PR #102, opened 2023-05-19
Pedro/bal 385 form base

yvesfracari:
https://www.loom.com/share/032ec8d7d0894ea1a4416264d4955f0f


packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -7,9 +7,9 @@ import { PoolAttribute } from "#/components/SearchPoolForm";
 import { pools } from "#/lib/gql";
 
 export interface TokensData {
-  symbol: string;
-  balance: number;
-  rate: number;
+  symbol: string | undefined;
+  balance: number | undefined;
+  rate: number | undefined;
```
ribeirojose, line 12:
```suggestion
  symbol?: string;
  balance?: number;
  rate?: number;
```

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    setInitialData(formData);
+  };
+
+  useEffect(() => {
+    setIsSubmitted(
+      typeof formData?.ampFactor == "number" &&
+        typeof formData?.swapFee == "number"
```
ribeirojose, line 49: If you want to set it as submitted just set it to true to keep it simple. If there's a reason for this it should probably have a comment around it since it's a very non obvious choice. Also likely to generate less [WTFs] imo (https://www.reddit.com/r/programming/comments/6mecg/wtfminutethe_only_valid_measurement_of_code/) (e.g. What if one of the types is not a number? Then it's not gonna be submitted and the user is gonna be stuck in a loop here? )

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+    );
+  }, [formData]);
```
ribeirojose, line 51: why would we be setting thing as submitted only when the formData changes and not when the `onSubmit` hook is called?

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+
+  useEffect(() => {
+    if (initialData) {
+      setFormData(initialData);
+    }
+  }, [initialData]);
```
ribeirojose, line 57: This is redundant with initializing the state with `initialData`

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+
+import Button from "#/components/Button";
+import { Input } from "#/components/Input";
+import { AnalysisData, useStableSwap } from "#/contexts/StableSwapContext";
 
 export default function PoolParametersForm() {
-  const { initialData } = useStableSwap();
+  const { initialData, setInitialData } = useStableSwap();
+  const [isSubmitted, setIsSubmitted] = useState(false);
+  const [formData, setFormData] = useState(initialData);
```
ribeirojose, line 12: do we even need this state here? Couldn't you just push everything to your context?

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+
+  const inputParameters = (field: keyof AnalysisData) => {
+    const label = field.replace(/([a-z])([A-Z])/g, "$1 $2").toLowerCase();
+    if (field == "tokens") return;
```
ribeirojose, line 16: 👀

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,76 @@
...
+    return {
+      label: label.charAt(0).toUpperCase() + label.slice(1),
```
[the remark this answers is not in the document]
yvesfracari: The js string doesn't have a capitalization function. In your opinion, should I create this function?

ribeirojose (approved):
🔝  just left a few suggestions

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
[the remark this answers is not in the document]
ribeirojose: Think someone added one in `packages/shared/index.ts`

ribeirojose (approved):
LGTM 👍 I left a few ideas in https://github.com/bleu-studio/balancer-pool-metadata/pull/104

packages/next-app/src/app/stableswapsimulator/(components)/PoolParametersForm.tsx
```diff
@@ -1,13 +1,87 @@
...
+          return;
+        }
+        if (value < 0) return;
+        setInitialData({
+          ...initialData,
+          [field]: value,
+        });
+      },
+    };
+  };
```
ribeirojose, line 47: curious to learn more about what value you had in mind to add with this abstraction -- I think this is more of a style preference, but I really dislike this pattern. Its abstraction adds indirection and is coupled with the implementation of your fields -- you're predefining  a placeholder pattern and a number-typed field, as well as with a state setter (`setInitialData`) and spreading the `register` from `react-hook-form` within it which I think could be bug prone since I don't think we could guarantee this function will be defined every time the component rerenders, and I don't know what are the expectations of react-hook-form about when it's gonna be called / by whom.

yvesfracari: I'll open a new issue to solve this. I checked the way that you implemented but when you import a pool is not change the fields.

luizakp (approved):

PR #103, opened 2023-05-19
[Balancer tools] Adapt colors to use only radixs palette

luizakp:
This PR adds:
- Eslint custom rule to not allow tailwind colors;
- Substitute to radix colors


packages/eslint-config/index.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     "tailwindcss/no-arbitrary-value": "off",
     "tailwindcss/no-custom-classname": "off",
     "tailwindcss/no-contradicting-classname": "error",
+    "custom-rules/no-tailwind-colors": "error",
```
ribeirojose, line 42: 😮

packages/eslint-custom-rules/no-tailwind-colors.js
```diff
@@ -0,0 +1,39 @@
...
+      schema: [],
+    },
+    create: function (context) {
+      const tailwindColors = [
+        'gray-', 'red-', 'yellow-', 'green-', 'blue-', 'indigo-', 'purple-', 'pink-',
+        'white-', 'black-',
+      ];
+  
+      function isTailwindColor(value) {
+        return tailwindColors.some((color) => value.startsWith(`text-${color}`) || value.startsWith(`bg-${color}`));
```
ribeirojose, line 18: not sure if this is the best approach because we'd also have border, shadow, etc 🤔

packages/eslint-custom-rules/no-tailwind-colors.js
```diff
@@ -0,0 +1,39 @@
+module.exports = {
```
ribeirojose, line 1: 👀 interesting suggestion!

packages/eslint-custom-rules/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "eslint-plugin-no-tailwind-colors",
```
ribeirojose, line 2: would consider adding this to the existing eslint-config package unless there's an explicit reason to add another one.

ribeirojose (approved):

ribeirojose (approved):

mendesfabio (approved):

packages/eslint-custom-rules/package.json
luizakp: tried to look more into this, but couldn't make it work from inside the eslint-config package. Do you have a suggestion?

PR #105, opened 2023-05-22
[StableSwap] Solve math import

luizakp:
This PR fix the stable functions import onto the Next App and remove moralis package

Example of import:

image


packages/math/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export * as metastable from "./metastable";
-export * as stable from "./stable";
+export * as Metastable from "./metastable";
```
yvesfracari, line 1: Maybe we could rename it as MetaStableMath and StableMath

yvesfracari (approved):
LGTM!

packages/math/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export * as metastable from "./metastable";
-export * as stable from "./stable";
+export * as MetaStableMath from "./metastable";
+export { ExtendedStableMath as StableMath } from "./stable";
```
ribeirojose, line 2: 👍 great refactor

PR #106, opened 2023-05-23
Pedro/bal 386 tokens table

yvesfracari:
This PR closes BAL-386, BAL-375, and BAL-407

https://www.loom.com/share/a84003bb30354218b7e5a5bf0bf7f229


packages/next-app/src/components/Table.tsx
```diff
@@ -103,7 +103,7 @@ function BodyCell({
     <td
       className={cn(
         "whitespace-nowrap text-sm text-gray-500",
-        customWidth ? cn(customWidth, "pl-4") : "p-4"
+        customWidth ? cn(customWidth, "pl-4") : "px-1 py-4"
```
ribeirojose, line 106: you can also just use custom width where you need -- remember this is going to affect all uses of the table component.

packages/schema/index.ts
```diff
@@ -140,3 +140,46 @@ export const AddressSchema = z.object({
...
+
+export const getStableSwapSimulatorTokensSchema = ({
+  symbolToEdit,
+  existentSymbols,
+}: {
+  symbolToEdit?: string;
+  existentSymbols?: string[];
+}) => {
+  const StableSwapSimulatorTokensSchema = z.object({
+    symbol: z
```
ribeirojose, line 152: wouldn't `z.enum(existentSymbols)` be enough? Also I think you'll lose the advantage that Zod would give you in generating the TS type if you pass `existentSymbols` as a string array. Not sure I fully get why this needs to be a standalone function -- are you planning on having multiple lists be attributable to this schema?

packages/next-app/src/components/Table.tsx
```diff
@@ -77,7 +77,7 @@ function HeaderCell({ children }: React.PropsWithChildren) {
   return (
     <th
       scope="col"
-      className="text-gray-40 p-4 text-left text-sm font-semibold"
+      className="text-gray-40 px-1 py-4 text-left text-sm font-semibold"
```
ribeirojose, line 80: You're changing the component everywhere -- just a reminder to ensure it looks nice across all uses.

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+    setInitialData({
+      ...initialData,
+      tokens: initialData?.tokens?.filter((token) => token.symbol !== symbol),
+    });
+  };
+
+  return (
+    <Table.BodyRow key={token.symbol}>
+      <Table.BodyCell>
+        <ButtonToOpenTokenForm
```
ribeirojose, line 82:
```suggestion
        <AddTokenButton
```

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Table.BodyCell>{token.symbol}</Table.BodyCell>
+      <Table.BodyCell>{token.balance?.toFixed()}</Table.BodyCell>
+      <Table.BodyCell>{token.rate?.toPrecision(2)}</Table.BodyCell>
+      <Table.BodyCell>
+        <button
+          type="button"
+          className="leading-none"
+          onClick={() => deleteToken(token?.symbol)}
+        >
+          <MinusCircledIcon
```
ribeirojose, line 102: This icon is not centralized horizontally -- I think you could just replicate what's in the metadata app here

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <Table.BodyRow key={token.symbol}>
+      <Table.BodyCell>
+        <ButtonToOpenTokenForm
+          icon={
+            <Pencil1Icon
+              width={22}
+              height={22}
+              className="text-amber9 hover:text-amber11"
+            />
```
ribeirojose, line 88: Also noticed in your video this is not vertically centralized -- think we could use the same pattern from the metadata table?

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <Dialog
+      title="Add token parameters"
+      content={<TokenForm symbolToEdit={symbolToEdit} />}
+    >
+      <button type="button" className="leading-none">
+        {icon}
+      </button>
+    </Dialog>
+  );
+}
```
ribeirojose, line 67: Does this really need to be another component?

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+}: React.PropsWithChildren<{
+  icon: React.ReactElement;
+  symbolToEdit?: string;
+}>) {
+  return (
+    <Dialog
+      title="Add token parameters"
+      content={<TokenForm symbolToEdit={symbolToEdit} />}
+    >
+      <button type="button" className="leading-none">
```
ribeirojose, line 62: wonder if `leading-none` is what's causing it to be non centralized 🤔

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Table>
+        <Table.HeaderRow>
+          <Table.HeaderCell>
+            <span className="sr-only">Edit</span>
+          </Table.HeaderCell>
+          <Table.HeaderCell>Symbol</Table.HeaderCell>
+          <Table.HeaderCell>Balance</Table.HeaderCell>
+          <Table.HeaderCell>Rate</Table.HeaderCell>
+          <Table.HeaderCell>
+            <ButtonToOpenTokenForm
```
ribeirojose, line 29: I think this icon is also not centralized vertically 🤔

packages/next-app/src/app/stableswapsimulator/(components)/TokenForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }
+    return true;
+  };
+
+  function addOrEditToken(data: TokensData) {
+    if (symbolToEdit) {
+      setInitialData({
+        ...initialData,
+        tokens: initialData?.tokens?.map((token: TokensData) => {
+          if (token.symbol == symbolToEdit) {
```
ribeirojose, line 55: not relevant in this case, but be mindful of the [strict equality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality)

ribeirojose, line 55: since this would always be `false` when `symbolToEdit` is not present you could really just have this condition, no?
```suggestion
        tokens: (initialData?.tokens || [])?.map((token: TokensData) => {
          if (token.symbol == symbolToEdit) {
```
but when would `initialData.tokens` be undefined/null? Shouldn't we initialize it with an empty array at least?

ribeirojose (approved):

packages/next-app/src/components/Table.tsx
yvesfracari: I will test it!

packages/next-app/src/app/stableswapsimulator/(components)/TokenTable.tsx
yvesfracari: I want to make sure that the Dialog will be the same in the edit and add buttons (just with the icon as different)

PR #107, opened 2023-05-24
[StableSwap] Add Analysis route

luizakp:
**NOTE: the large amount of added lines is due to pnpm-lock.yaml**

This PR adds:
- analysis route + handle on Submit to send to this route;
- Spinner named export
- add packages math, sor and plotly
- handle new InitialState

https://www.loom.com/share/a95eaff0aacf4b18b6c5bccd235da8ed


packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -20,19 +20,30 @@ export interface AnalysisData {
 
 interface StableSwapContextType {
   initialData: AnalysisData;
+  newData?: AnalysisData;
```
ribeirojose, line 23: IMO this distinction between initial/new data is a bit confusing. I asked ChatGPT for some ideas and here's a few:

> Original vs Updated: This terminology clearly signifies that the first set of parameters was the initial or original one, and the second set represents a change or an update.
> 
> Before vs After: This pair emphasizes the temporal order of the parameter sets, making it clear that the "Before" set was entered first, and the "After" set represents a change.
> 
> Primary vs Secondary: This pair avoids connotations of time or change, instead simply differentiating between the two sets of parameters. It still implies an order, with the Primary set being more important or entered first.
> 
> Start vs End: Similar to "Before vs After," but uses a different metaphor, implying a process or journey from the Start parameters to the End ones.
> 
> Input vs Comparison: This would imply that the "Input" parameters are the ones that the user is mainly interested in, and the "Comparison" parameters are being used to check, verify, or contrast against them.
> 
> Baseline vs Variant: This could imply that the Baseline parameters are the standard or default ones, and the Variant parameters represent a change or deviation from that standard.
> 
> Reference vs Test: This pair implies that the "Reference" parameters are the known, trusted, or verified ones, and the "Test" parameters are being evaluated against them.
> 
> Control vs Experimental: Borrowed from experimental sciences, this pair of terms could be useful if the app is often used for testing changes or innovations.
> 

I think the three last ones make more sense to our case, but wanted to paste the full output so that you maybe have any other ideas!

packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -20,19 +20,30 @@ export interface AnalysisData {
...
+  indexAnalysisToken: number;
+  indexCurrentTabToken: number;
+  setIndexAnalysisToken: (index: number) => void;
+  setIndexCurrentTabToken: (index: number) => void;
   setInitialData: (data: AnalysisData) => void;
   handleImportPoolParametersById: (data: PoolAttribute) => void;
   newPoolImportedFlag: boolean;
+  defaultInitialData: AnalysisData;
```
ribeirojose, line 31: Why would this need to be passed down to consumers?

packages/next-app/src/app/stableswapsimulator/page.tsx
```diff
@@ -1,7 +1,15 @@
...
+
+import { useStableSwap } from "#/contexts/StableSwapContext";
+
 import ParametersNotSet from "./(components)/ParametersNotSet";
 
 export default function Page() {
+  const { setInitialData, defaultInitialData } = useStableSwap();
+  useEffect(() => {
+    setInitialData(defaultInitialData);
+  }, []);
```
ribeirojose, line 13: Shouldn't this be called from within `useStableSwap` or shouldn't we have the default initial value as this?

packages/next-app/src/app/stableswapsimulator/(components)/GraphView.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import StableCurve from "./StableCurve";
+
+export function GraphView() {
+  const { initialData } = useStableSwap();
+  const { push } = useRouter();
+
+  if (!initialData.swapFee) {
+    push("/stableswapsimulator");
+  }
```
ribeirojose, line 15: I think I `#would` move this `push` to within the React Context since it's not really a composable component, just somewhere we've decided to concentrate state logic that's related to stableswap.

ribeirojose (approved):

packages/next-app/src/app/stableswapsimulator/(components)/InitialDataForm.tsx
```diff
@@ -73,7 +77,7 @@ export default function InitialDataForm() {
         )}
         <div hidden={true}>
           {initialData?.tokens.map((token, i) => (
-            <>
+            <div key={token.symbol}>
```
yvesfracari, line 80: 👍

packages/next-app/src/contexts/StableSwapContext.tsx
luizakp: Baseline and Variant sounds good

yvesfracari (approved):
LGTM

yvesfracari (approved):
LGTM

PR #108, opened 2023-05-24
Pedro/bal 383 sidebar with controllers

yvesfracari:
https://www.loom.com/share/69caa8f768be44ffb5c0caf97be40f27

### DIFF
--- a/packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
@@ -48,13 +48,13 @@ export default function BaselineDataForm() {
     };
   };
 
-  const handleInputChange = () => {
+  const handleChange = () => {
     if (Object.keys(errors).length) return;
     const data = getValues();
     setBaselineData(data as AnalysisData);
   };
 
-  useEffect(handleInputChange, [debouncedSwapFee, debouncedAmpFactor, tokens]);
+  useEffect(handleChange, [debouncedSwapFee, debouncedAmpFactor, tokens]);
 
   useEffect(clearErrors, [baselineData?.tokens, newPoolImportedFlag]);
 
--- a/packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
@@ -1,0 +1,99 @@
+"use client";
+
+import { StableSwapSimulatorDataSchema } from "@balancer-pool-metadata/schema";
+import { capitalize } from "@balancer-pool-metadata/shared";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { useEffect } from "react";
+import { useForm } from "react-hook-form";
+
+import { Input } from "#/components/Input";
+import { AnalysisData, useStableSwap } from "#/contexts/StableSwapContext";
+import useDebounce from "#/hooks/useDebounce";
+
+import { TokenTable } from "./TokenTable";
+
+export default function BaselineDataForm() {
+  const { baselineData, setBaselineData, newPoolImportedFlag } =
+    useStableSwap();
+  const {
+    register,
+    getValues,
+    clearErrors,
+    watch,
+    formState: { errors },
+  } = useForm<typeof StableSwapSimulatorDataSchema._type>({
+    resolver: zodResolver(StableSwapSimulatorDataSchema),
+    mode: "onChange",
+  });
+
+  const swapFee = watch("swapFee");
+  const ampFactor = watch("ampFactor");
+  const tokens = watch("tokens");
+  const debouncedSwapFee = useDebounce(swapFee);
+  const debouncedAmpFactor = useDebounce(ampFactor);
+
+  const inputParameters = (field: keyof AnalysisData) => {
+    const label = field.replace(/([a-z])([A-Z])/g, "$1 $2").toLowerCase();
+    if (field == "tokens") return;
+    return {
+      ...register(field, {
+        required: true,
+        value: baselineData?.[field],
+        valueAsNumber: true,
+      }),
+      label: capitalize(label),
+      placeholder: `Define the initial ${label}`,
+      errorMessage: errors[field]?.message?.toString() || "",
+      form: "baseline-data-form",
+    };
+  };
+
+  const handleInputChange = () => {
+    if (Object.keys(errors).length) return;
+    const data = getValues();
+    setBaselineData(data as AnalysisData);
+  };
+
+  useEffect(handleInputChange, [debouncedSwapFee, debouncedAmpFactor, tokens]);
+
+  useEffect(clearErrors, [baselineData?.tokens, newPoolImportedFlag]);
+
+  return (
+    <div className="flex flex-col gap-4">
+      <form id="baseline-data-form" />
+      <Input {...inputParameters("swapFee")} />
+      <Input {...inputParameters("ampFactor")} />
+      <div className="flex flex-col">
+        <label className="mb-2 block text-sm text-slate12">Tokens</label>
+        {errors?.tokens?.message && (
+          <div className="h-6 mt-1 text-tomato10 text-sm">
+            <span>{errors?.tokens?.message}</span>
+          </div>
+        )}
+        <div hidden={true}>
+          {baselineData?.tokens.map((token, i) => (
+            <div key={token.symbol}>
+              <input
+                form="baseline-data-form"
+                {...register(`tokens.${i}.symbol`, {
+                  value: token.symbol,
+                })}
+              />
+              <input
+                form="baseline-data-form"
+                {...register(`tokens.${i}.balance`, { value: token.balance })}
+              />
+              <input
+                form="baseline-data-form"
+                {...register(`tokens.${i}.rate`, {
+                  value: token.rate,
+                })}
+              />
+            </div>
+          ))}
+        </div>
+        <TokenTable />
+      </div>
+    </div>
+  );
+}
--- a/packages/next-app/src/app/stableswapsimulator/(components)/Menu.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/Menu.tsx
@@ -1,0 +1,88 @@
+import { MagnifyingGlassIcon } from "@radix-ui/react-icons";
+import * as Separator from "@radix-ui/react-separator";
+import * as Tabs from "@radix-ui/react-tabs";
+import { usePathname } from "next/navigation";
+
+import Sidebar from "#/components/Sidebar";
+
+import BaselineDataForm from "./BaselineDataForm";
+import InitialDataForm from "./InitialDataForm";
+import { SearchPoolFormDialog } from "./SearchPoolFormDialog";
+import VariantDataForm from "./VariantDataForm";
+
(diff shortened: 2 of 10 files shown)


packages/next-app/src/components/Select.tsx
```diff
@@ -30,7 +30,7 @@ export const Select = forwardRef<HTMLButtonElement, ISelect>(
       <SelectPrimitive.Root {...props}>
         <SelectPrimitive.Trigger
           className={cn(
-            "inline-flex h-[35px] outline-none items-center justify-center gap-[5px] rounded px-[15px] text-[13px] leading-none ",
+            "inline-flex h-[35px] outline-none items-center justify-between gap-[5px] rounded px-[15px] text-[13px] leading-none ",
```
luizakp, line 33: just checked here on metadata and input is looking fine there too hhahahha

luizakp (approved):
TS code looking great 😄 just left some comments

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Input {...inputParameters("swapFee")} />
+      <Input {...inputParameters("ampFactor")} />
+      <div className="flex flex-col">
+        <label className="mb-2 block text-sm text-slate12">Tokens</label>
+        {errors?.tokens?.message && (
+          <div className="h-6 mt-1 text-tomato10 text-sm">
+            <span>{errors?.tokens?.message}</span>
+          </div>
+        )}
+        <div hidden={true}>
```
devjoaov, line 73:
```suggestion
        <div hidden>
```

packages/next-app/src/components/Sidebar.tsx
```diff
@@ -25,14 +25,16 @@ function useSidebarContext() {
 export default function Sidebar({
   children,
   isFloating = false,
+  padding = "py-5 px-4",
```
ribeirojose, line 28: You could also give a name more true to what the function of this variable is and simply call this `classNames` and add it to the end of the `cn`.

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  useEffect(() => {
+    if (variantData?.swapFee) setValue("swapFee", variantData?.swapFee);
+    if (variantData?.ampFactor) setValue("ampFactor", variantData?.ampFactor);
+  }, [variantData]);
+
+  return (
+    <div className="flex flex-col gap-4">
+      <form id="variant-data-form" />
```
ribeirojose, line 44: This is the equivalent of `<form></form>`. Is that what you meant to do here?

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+    setBaselineData(data as AnalysisData);
+  };
+
+  useEffect(handleChange, [debouncedSwapFee, debouncedAmpFactor, tokens]);
+
+  useEffect(clearErrors, [baselineData?.tokens, newPoolImportedFlag]);
+
+  return (
+    <div className="flex flex-col gap-4">
+      <form id="baseline-data-form" />
```
ribeirojose, line 63: Not sure if this is what you were envisioning, this is just the equivalent of `<form></form>`

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const handleChange = () => {
+    if (Object.keys(errors).length) return;
+    const data = getValues();
+    setVariantData(data as AnalysisData);
+  };
+
+  useEffect(() => {
+    if (variantData?.swapFee) setValue("swapFee", variantData?.swapFee);
+    if (variantData?.ampFactor) setValue("ampFactor", variantData?.ampFactor);
+  }, [variantData]);
```
ribeirojose, line 40: Not sure I follow this, can you elaborate? If you want to set a default value you can add that to the `useForm` hook

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  } = useForm<typeof StableSwapSimulatorDataSchema._type>({
+    resolver: zodResolver(StableSwapSimulatorDataSchema),
+    mode: "onChange",
+  });
+
+  const handleChange = () => {
+    if (Object.keys(errors).length) return;
+    const data = getValues();
+    setVariantData(data as AnalysisData);
+  };
```
ribeirojose, line 35: Shouldn't we just handle onSubmit as the user hits enter/clicks to refresh the simulation here?

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+          Analysis Token
+        </label>
+        <Select
+          onValueChange={(i) => setIndexAnalysisToken(Number(i))}
+          defaultValue={indexAnalysisToken.toString()}
+        >
+          {baselineData?.tokens.map(({ symbol }, index) => (
+            <SelectItem value={index.toString()}>{symbol}</SelectItem>
+          ))}
+        </Select>
```
ribeirojose, line 56: Couldn't we simply use the symbol everywhere?

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+          required: true,
+          value: variantData?.swapFee,
+          valueAsNumber: true,
+          onChange: handleChange,
+        })}
+        label="Swap fee"
+        rightLabel={`baseline: ${baselineData?.swapFee}`}
+        placeholder="Define the new swap fee"
+        errorMessage={errors?.swapFee?.message}
+        form="new-data-form"
```
ribeirojose, line 69:
```suggestion
        form="variant-data"
```

packages/next-app/src/app/stableswapsimulator/(components)/Menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+  );
+}
+
+export default function Menu() {
+  const pathname = usePathname();
+  if (pathname.includes("/analysis")) {
+    return <AnalysisMenu />;
+  }
+  return <IndexMenu />;
+}
```
ribeirojose, line 88: Not sure I follow the purpose of this check since these are two completely distinct components. What do we gain by adding this here instead of just leaving each component at their respective page?

packages/next-app/src/app/stableswapsimulator/(components)/Menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+        defaultValue="variantData"
+      >
+        <Tabs.List className="shrink-0 flex border-b border-blue1 bg-blue3">
+          <Tabs.Trigger
+            className="px-5 h-[45px] flex-1 flex items-center justify-center text-[15px] hover:text-white hover:bg-blue4 data-[state=active]:text-white data-[state=active]:shadow-[inset_0_-1px_0_0,0_1px_0_0] data-[state=active]:gray-800 data-[state=active]:focus:shadow-[0_0_0_2px] data-[state=active]:focus:shadow:gray-800 outline-none cursor-default"
+            value="baselineData"
+          >
+            <div className="flex justify-center items-center h-screen gap-x-3">
+              <div className="w-3 h-3 rounded-full bg-blue-700" />
+              Initial
```
ribeirojose, line 27: This should probably be `Baseline` right?

packages/next-app/src/app/stableswapsimulator/(components)/Menu.tsx
ribeirojose: Another comment here is that we could just have `AnalysisMenu` on all pages and simplify things -- also perhaps `SearchPoolFormDialog` could be always displayed in the first tab.

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  } = useForm<typeof StableSwapSimulatorDataSchema._type>({
+    resolver: zodResolver(StableSwapSimulatorDataSchema),
+    mode: "onChange",
+  });
+
+  const swapFee = watch("swapFee");
+  const ampFactor = watch("ampFactor");
+  const tokens = watch("tokens");
+  const debouncedSwapFee = useDebounce(swapFee);
+  const debouncedAmpFactor = useDebounce(ampFactor);
```
ribeirojose, line 33: It almost looks this defeats the purpose of having a form in the first place. Same with the setting of params we do in the context through `handleChange`. A much more elegant solution IMO would be triggering the form submission remotely -- e.g. define `const submit = useDebounce(handleSubmit(onSubmit))` and call `submit()` when you want to submit it. It leverages more of the form logic instead of reimplementing it

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
yvesfracari: Think that we didn't discuss this that much but my idea was the simulation refresh on change of each input. What do you think about it?

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const inputParameters = (field: keyof AnalysisData) => {
```
ribeirojose, line 35: Still think this is bug-prone and obfuscates what's going on inside the form

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
ribeirojose, line 73: what's the purpose of this hidden div here again?

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
ribeirojose: If that's not computationally intensive I think it would work -- just wonder if it won't be to jittery -- but I think you added the debounced values because of this. Personally I'd not reimplement the form logic and just call submit (debounced, perhaps) after an input changes and I wanted the simulations to be refreshed.

packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
yvesfracari: The math packages expected an index, so we need to use the index as the reference

packages/next-app/src/app/stableswapsimulator/(components)/BaselineDataForm.tsx
yvesfracari: I also thought to remove the form (since we are dealing with changes as soon as they happen) but I want to reuse our validations. Will look for the solution that we described.

PR #109, opened 2023-05-25
[Stable Swap] Swap curve for baseline data

luizakp:
[Linear Issue](https://linear.app/bleu-llc/issue/BAL-419/swap-curve-for-baseline-data)

This PR adds:
- Swap Curve for the baseline data;
- Custom hover;
- Loading state for when the graph is not ready;
- Clear Baseline Form (not only the state) when the user is back on the homepage

https://www.loom.com/share/7a12f81a4bc94978b51decf635505b59

### DIFF
--- a/packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/VariantDataForm.tsx
@@ -44,7 +44,9 @@ export default function NewDataForm() {
           defaultValue={indexAnalysisToken.toString()}
         >
           {baselineData?.tokens.map(({ symbol }, index) => (
-            <SelectItem value={index.toString()}>{symbol}</SelectItem>
+            <SelectItem key={symbol} value={index.toString()}>
+              {symbol}
+            </SelectItem>
           ))}
         </Select>
       </div>


packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -78,16 +102,69 @@ export function StableSwapProvider({ children }: PropsWithChildren) {
...
+      allBalances: allBalancesOldBn,
+      allBalancesScaled: allBalancesBn,
+      amp: numberToBigNumber(amp, 3),
+      tokenIndexIn: indexIn,
+      tokenIndexOut: indexOut,
+      decimalsIn: 6,
+      decimalsOut: 18,
+    } as StablePoolPairData;
+  }
+
```
luizakp, line 150: We create this because the pools can actually exist if we import, but they can be made up and the StableMath expects data from a real pool

packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -35,6 +41,22 @@ interface StableSwapContextType {
...
+    allBalances,
+    amp,
+  }: {
+    indexIn: number;
+    indexOut: number;
+    swapFee: number;
+    allBalances: number[];
+    amp: number;
+  }) => StablePoolPairData;
+  numberToOldBigNumber: (number: number, decimals?: number) => typeof bnum;
```
yvesfracari, line 59: maybe we could export all the number and big number converts (opposite direction too)

packages/next-app/src/app/stableswapsimulator/(components)/StableCurve.tsx
```diff
@@ -1,32 +1,164 @@
...
+    <div className="text-white">
+      <Plot
+        data={[
+          {
+            x: initialAmountsAnalysisTokenIn,
+            y: initialAmountTabTokenOut,
+            type: "scatter",
+            mode: "lines",
+            marker: { color: blueDarkA.blueA9 },
+            hovertemplate: `Swap $%{x:.2f} ${tokensSymbol[indexAnalysisToken]} for ${formattedPoints?.y} ${tokensSymbol[indexCurrentTabToken]} <extra></extra>`,
```
yvesfracari, line 89: we should think of a way to round those values (depending on the scale). E.g. 10100.321-> 10.1k, 13.32->13.32.

packages/next-app/src/app/stableswapsimulator/(components)/StableCurve.tsx
```diff
@@ -1,32 +1,164 @@
...
+          {
+            x: initialAmountsAnalysisTokenOut,
+            y: initialAmountTabTokenIn,
+            type: "scatter",
+            mode: "lines",
+            marker: { color: blueDarkA.blueA9 },
+            hovertemplate: `Swap $%{y:.2f} ${tokensSymbol[indexCurrentTabToken]} for ${formattedPoints?.x} ${tokensSymbol[indexAnalysisToken]}<extra></extra>`,
+            showlegend: false,
+          },
+        ]}
```
yvesfracari, line 101: would add both markers in the same trace and show the legend, because the user can click on it to hide it for example.

yvesfracari (approved):

packages/next-app/src/app/stableswapsimulator/(components)/StableCurve.tsx
```diff
@@ -1,32 +1,164 @@
...
+    numberToOldBigNumber,
   } = useStableSwap();
-  const amountsAnalysisToken = calculateAmounts(
-    baselineData?.tokens?.[indexAnalysisToken]?.balance
+
+  if (
+    !baselineData ||
+    !baselineData.swapFee ||
+    !baselineData.ampFactor ||
+    !baselineData.tokens
```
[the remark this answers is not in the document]
luizakp: This is also to make sure that the `baselineData.swapFee` and `baselineData.ampFactor` is not undefined, which is needed for the `preparePoolPairData` function. If I remove this we get a type error

packages/next-app/src/app/stableswapsimulator/(components)/StableCurve.tsx
luizakp: [Issue for this](https://linear.app/bleu-llc/issue/BAL-421/hover-decimals-depending-of-pool-scale)

packages/next-app/src/app/stableswapsimulator/(components)/StableCurve.tsx
luizakp: cool, can we update the figma then? or specify this on linear? Because I just followed the proposed directions from both

PR #115, opened 2023-05-29
[Internal Manager] Use Radix colors only

luizakp:


packages/next-app/src/components/Table.tsx
```diff
@@ -80,7 +80,7 @@ function HeaderCell({
   return (
     <th
       scope="col"
-      className={cn("text-gray-40 text-left text-sm font-semibold", padding)}
+      className={cn("text-gray12 text-left text-sm font-semibold", padding)}
```
ribeirojose, line 83: I think we're using slate instead of bare gray

ribeirojose (dismissed):

ribeirojose (approved):

PR #117, opened 2023-05-29
[Stableswap Simulator] add generic Plot component with default layout and config

yvesfracari:
Use example:

image

### DIFF
--- a/packages/next-app/src/components/Plot.tsx
+++ b/packages/next-app/src/components/Plot.tsx
@@ -1,6 +1,14 @@
 import { amberDarkA, blueDarkA, grayDarkA } from "@radix-ui/colors";
 import { merge } from "lodash";
-import PlotRoot, { PlotParams } from "react-plotly.js";
+import dynamic from "next/dynamic";
+import { PlotParams } from "react-plotly.js";
+
+import { Spinner } from "#/components/Spinner";
+
+const PlotRoot = dynamic(() => import("react-plotly.js"), {
+  ssr: false,
+  loading: () => <Spinner />,
+});
 
 export const defaultAxisLayout = {
   gridcolor: grayDarkA.grayA10,
--- a/packages/next-app/src/components/Plot.tsx
+++ b/packages/next-app/src/components/Plot.tsx
@@ -1,0 +1,47 @@
+import { amberDarkA, blueDarkA, grayDarkA } from "@radix-ui/colors";
+import { merge } from "lodash";
+import PlotRoot, { PlotParams } from "react-plotly.js";
+
+export const defaultAxisLayout = {
+  gridcolor: grayDarkA.grayA10,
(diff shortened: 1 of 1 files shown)


packages/next-app/src/components/Plot.tsx
```diff
@@ -0,0 +1,47 @@
...
+  useResizeHandler: true,
+  layout: {
+    plot_bgcolor: blueDarkA.blueA1,
+    paper_bgcolor: blueDarkA.blueA1,
+    font: {
+      color: grayDarkA.grayA12,
+      family: "Inter",
+    },
+    xaxis: JSON.parse(JSON.stringify(defaultAxisLayout)),
+    yaxis: JSON.parse(JSON.stringify(defaultAxisLayout)),
```
ribeirojose, line 24: 👀 care to enlighten me in why would we need this here?

packages/next-app/src/components/Plot.tsx
```diff
@@ -0,0 +1,47 @@
...
+    showEditInChartStudio: false,
+    showLink: false,
+    watermark: false,
+  },
+  revision: 0,
+};
+
+export default function Plot(props: PlotParams) {
+  const plotProps = merge(JSON.parse(JSON.stringify(defaultPlotProps)), props);
+  return <PlotRoot {...plotProps} key={JSON.stringify(plotProps)} />;
```
ribeirojose, line 46: I'd suggest letting the caller define a key here -- probably not an issue now but plots with the same props would probably lead to a bug in how plotly is handling these plots

packages/next-app/src/components/Plot.tsx
```diff
@@ -0,0 +1,47 @@
...
+    showSendToCloud: false,
+    showEditInChartStudio: false,
+    showLink: false,
+    watermark: false,
+  },
+  revision: 0,
+};
+
+export default function Plot(props: PlotParams) {
+  const plotProps = merge(JSON.parse(JSON.stringify(defaultPlotProps)), props);
```
ribeirojose, line 45: this is very odd and prone to WTFs by newcomers -- if we must do it please leave a comment here so whomever understands why this is needed

ribeirojose (changes requested):
only `#must` is address the subsequent json.parse/json.stringify calls. That smells like it's covering into another issue and can very quickly lead to spaghetti code. We must add a comment explaining it or fix whatever the underlying issue is.

packages/next-app/src/components/Plot.tsx
yvesfracari: This is needed bc the merge changes the first object (defaultPlotProps) with the function result. So, I am using the json.parse/json.stringify to deep copy the defaultPlotProps

ribeirojose (dismissed):

PR #118, opened 2023-05-30
Pedro/bal 388 depth cost chart

yvesfracari:
This PR adds:
- Fix variant tokens updates
- Add depth cost chart (on video)
- Define layout of graphView component (overflow and flex)

https://www.loom.com/share/5c5d3e4895db47458afe25054f175906

### DIFF
--- a/packages/math/src/stable/index.ts
+++ b/packages/math/src/stable/index.ts
@@ -162,6 +162,45 @@ function _tokenInForExactSpotPriceAfterSwap(
   );
 }
 
+function _tokenOutForExactSpotPriceAfterSwap(
+  spotPrice: OldBigNumber,
+  poolPairData: StablePoolPairData,
+  outGuess = poolPairData.allBalances[poolPairData.tokenIndexOut].times(
+    bnum(0.01)
+  ),
+  spotPricePrecision = bnum(0.0000001),
+  iteration_number = 0
+): OldBigNumber {
+  // Calculate the amount of tokenIn needed to reach the desired spotPrice
+  // The Newton-Raphson method is used to find the SpotPrice of the function
+  // Results could be inaccurate for very small amounts of tokenIn
+  const guessedSpotPrice =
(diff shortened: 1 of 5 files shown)


packages/math/src/stable/index.ts
```diff
@@ -162,6 +162,45 @@ function _tokenInForExactSpotPriceAfterSwap(
...
+    StableMaths._spotPriceAfterSwapTokenInForExactTokenOut(outGuess, {
+      ...poolPairData,
+    });
+  const diffFromSpotPrice = spotPrice.minus(guessedSpotPrice);
+  if (diffFromSpotPrice.abs().lte(spotPricePrecision)) {
+    return outGuess;
+  }
+
+  iteration_number += 1;
+  if (iteration_number > 255) {
```
luizakp, line 187: how did you define 255?

packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { StableMath } from "@balancer-pool-metadata/math/src";
+import { PlotType } from "plotly.js";
+
+import Plot, { defaultAxisLayout } from "#/components/Plot";
+import { AnalysisData, useStableSwap } from "#/contexts/StableSwapContext";
+
+export default function DepthCost() {
+  const { indexAnalysisToken, baselineData, variantData } = useStableSwap();
+
+  const tokenY = baselineData?.tokens[indexAnalysisToken];
```
luizakp, line 13: hmmm I think we should keep the same convention for names and objects data to be consumed by the plot. If we decide this still we should refactor the other charts

luizakp (approved):
just left some comments and a open thought hehehe

packages/math/src/stable/index.ts
yvesfracari: It is the same number to calculate the invariant in the solidity

ribeirojose: `#should` a comment here it just pointing that out 😬

PR #119, opened 2023-05-30
Pedro/bal 431 handle metastable conversions inside math package

yvesfracari:
This PR:
- Remove math related functions from context;
- Combine MetaStable conversion with MetaStable;
- Refactor exports of stable math;

### DIFF
--- a/packages/math/src/index.ts
+++ b/packages/math/src/index.ts
@@ -1,2 +1,146 @@
-export * as MetaStableMath from "./metastable";
-export { ExtendedStableMath as StableMath } from "./stable";
+import { bnum, OldBigNumber, StablePool } from "@balancer-labs/sor";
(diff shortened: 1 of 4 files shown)


packages/math/src/index.ts
```diff
@@ -1,2 +1,168 @@
...
+
+  return {
+    id: "0x",
+    address: "0x",
+    poolType: 1,
+    tokenIn: "0x",
+    tokenOut: "0x",
+    balanceIn: numberToBigNumber(balances[indexIn])
+      .mul(numberToBigNumber(rates[indexIn]))
+      .div(ONE),
```
luizakp, line 55: Why do we need to divide by WeiPerEther?

luizakp (approved):

PR #126, opened 2023-06-01
[Stable Swap] Add token decimals to consideration

luizakp:
This PR adds token decimals into consideration when creating the BigNumbers used on Math packages


luizakp:
Almost sure this doesn't actually affect any results because this only changes the amount of decimals after the comma 🤔

packages/next-app/src/app/stableswapsimulator/(components)/ImpactCurve.tsx
```diff
@@ -50,14 +52,15 @@ export function ImpactCurve() {
...
+      balances,
       amp,
+      decimals,
     });
 
     const impactTabTokenOut = amountsIn.map(
       (amount) =>
         StableMath._priceImpactForExactTokenInSwap(
-          numberToOldBigNumber(amount),
+          numberToOldBigNumber({ number: amount, decimals: decimals[indexIn] }),
```
mendesfabio, line 63: I'm not completely sure this is needed because these functions (e.g. [_exactTokenInForTokenOut](https://github.com/balancer/balancer-sor/blob/7b087e32cb9d3d24af301829b169ac88e67f1b54/src/pools/metaStablePool/metaStablePool.ts#LL204C5-L204C29)) from SDK do parse internally the numbers - did you found any issues not scaling yourself the number?

luizakp: to create those Big Numbers we used `parseFixed` or `toFixed` (oldBNum) which needed decimals. At first, we just considered everything with decimals = 18, which we thought was an inappropriate  approach since we have access to the real pair decimals

packages/schema/index.ts
```diff
@@ -168,6 +168,7 @@ export const getStableSwapSimulatorTokensSchema = ({
       ),
     balance: z.number().positive(),
     rate: z.number().positive(),
+    decimal: z.number().positive(),
```
yvesfracari, line 171:
```suggestion
    decimal: z.number().int().positive().max(60),
```

yvesfracari (dismissed):

packages/next-app/src/app/stableswapsimulator/(components)/ImpactCurve.tsx
luizakp: removed this for now https://github.com/bleu-studio/balancer-pool-metadata/pull/126/commits/9b2c4ee6e534ab9e060a3a076d70793a462c0627

mendesfabio (approved):

PR #127, opened 2023-06-02
[Stable Swap] Handle metastable conversions inside math package

luizakp:


packages/math/src/stable/index.ts
```diff
@@ -177,40 +177,22 @@ function _tokenInForExactSpotPriceAfterSwap(
...
-  outGuess = poolPairData.allBalances[poolPairData.tokenIndexOut].times(
+  firstGuess = poolPairData.allBalances[poolPairData.tokenIndexOut].times(
     bnum(0.01)
   ),
-  spotPricePrecision = bnum(0.0000001),
-  iteration_number = 0
+  spotPricePrecision = bnum(0.0000001)
 ): OldBigNumber {
-  // Calculate the amount of tokenIn needed to reach the desired spotPrice
+  // Calculate the amount of tokenOut needed to reach the desired spotPrice
```
ribeirojose, line 185: 👍

packages/math/src/index.ts
```diff
@@ -1,2 +1,261 @@
...
+): OldBigNumber {
+  const effectivePriceMetastable = effectivePriceForExactTokenOutSwap(
+    amountOut,
+    poolPairData
+  );
+  const spotPriceMetastable = spotPrice(poolPairData);
+  return bnum(1).minus(spotPriceMetastable.div(effectivePriceMetastable));
+}
+
+export const MetaStableMath = {
```
ribeirojose, line 250: `#note` we should probably rename this package to something else 😝

ribeirojose (approved):

packages/math/src/index.ts
luizakp: kkkkkk ok

PR #130, opened 2023-06-07
Pedro/bal 436 add swap simulation into graph view

yvesfracari:
This PR:
- Fix explanation of depth cost (price change vs price impact)
- Refactor analysis view page
- Add swap simulation

https://www.loom.com/share/4b7e90ade4ae456295e9774fdf9c00b2


packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
```diff
@@ -117,9 +117,9 @@ export function DepthCost() {
 
   const props = {
     data: data,
-    title: "Depth cost (-/+ 2% of price impact)",
+    title: "Depth cost (-/+ 2% of price change)",
```
luizakp, line 120: We were getting the concept wrong? ;-;

packages/next-app/src/app/stableswapsimulator/(components)/SwapSimulator.tsx
```diff
@@ -0,0 +1,258 @@
...
+    amp: data?.ampFactor,
+    rates: data?.tokens.map((token) => token.rate),
+    decimals: data?.tokens.map((token) => token.decimal),
+  });
+
+  const OldBigNumberAmount = MetaStableMath.numberToOldBigNumber(
+    amount
+  ) as OldBigNumber;
+
+  let amountIn = OldBigNumberAmount;
```
luizakp, line 222: We are using `let` because you want to reassign this variable without causing a re-render? I as this because I feel we avoid using `let` but it might be the behavior we want in this case

packages/next-app/src/app/stableswapsimulator/(components)/TokensDistribution.tsx
```diff
@@ -27,23 +27,21 @@ const colors = [
 export function TokensDistribution() {
   const { initialData, customData } = useStableSwap();
   return (
-    <div className="flex flex-col w-1/3 gap-y-2 text-slate12">
+    <div className="flex flex-col w-1/4 gap-y-4 text-slate12">
```
luizakp, line 30: Maybe the parent component has to have a max width and the charts have to work with the space they have because on smaller screens (eg. with sidebar from arc open) we get x-scroll

image

luizakp (approved):
great stuff!

packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
yvesfracari: Yes, were wrong. The depth cost calculation is based on spot price. So, it is how much you can add to a token to move the price to -/+ 2%. The price impact is effective price / spot price.

packages/next-app/src/app/stableswapsimulator/(components)/TokensDistribution.tsx
yvesfracari: humm! Lets open a issue to check this out

PR #133, opened 2023-06-09
Pedro/bal 435 add more points into swap impact curves

yvesfracari:
This PR:
- Block zero swap fee (to avoid math errors)
- Add more points to the curves
- Fix sign of curves calculations
- Improve tooltip description on curves (add sign description)


luizakp:
image

This first pool has a strange outcome

image

packages/next-app/src/app/stableswapsimulator/(components)/ImpactCurve.tsx
```diff
@@ -72,9 +72,17 @@ export function ImpactCurve() {
...
+        MetaStableMath.exactTokenInForTokenOut(
+          MetaStableMath.numberToOldBigNumber(amount),
+          poolPairDataOut
+        ).toNumber() * -1
+    );
+
     const impactTabTokenIn = amountsIn.map(
       (amount) =>
-        MetaStableMath.priceImpactForExactTokenOutReversedSwap(
+        MetaStableMath.priceImpactForExactTokenInReversedSwap(
```
luizakp, line 85: in this case, we have the "out" that is the `amountsIn` and we want to get the price Impact from that, so isn't the correct function?

image

luizakp (dismissed):

packages/next-app/src/app/stableswapsimulator/(components)/ImpactCurve.tsx
yvesfracari: Note:
- Improve code readability using 2 times the calculated amounts (one for token in and another one for out). Improve names as well.

PR #138, opened 2023-06-13
Pedro/bal 379 ability to share a link to a simulation

yvesfracari:
This PR creates a share simulation button. 

Limitation: This button just shares the current status of the dashboard, if something changes after the link is copied, the previously shared link will not have the changes.

https://www.loom.com/share/232460d53cfa48b6883ef4bf5f03e1f1

### DIFF
--- a/packages/next-app/src/app/stableswapsimulator/(components)/ShareButton.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/ShareButton.tsx
@@ -42,20 +42,24 @@ export function ShareButton() {
         </Button>
       </Tooltip>
       <Toast
-        content={
-          <div className="flex h-14 flex-row items-center justify-between px-4 py-8">
-            <div className="flex flex-col justify-between space-y-1">
-              <h1 className="text-xl font-medium text-slate12">Link copied</h1>
-              <h3 className="mb-2 text-sm leading-3 text-slate11">
-                Use it to access the dashboard state.
-              </h3>
-            </div>
-          </div>
-        }
+        content={<ToastContent />}
         isOpen={isNotifierOpen}
         setIsOpen={setIsNotifierOpen}
         duration={5000}
       />
     </>
+  );
+}
+
+function ToastContent() {
+  return (
+    <div className="flex h-14 flex-row items-center justify-between px-4 py-8">
+      <div className="flex flex-col justify-between space-y-1">
+        <h1 className="text-xl font-medium text-slate12">Link copied</h1>
+        <h3 className="mb-2 text-sm leading-3 text-slate11">
+          Use it to access the dashboard state.
+        </h3>
+      </div>
+    </div>
   );
 }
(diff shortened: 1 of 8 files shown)


packages/next-app/src/app/stableswapsimulator/(components)/ShareButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useStableSwap } from "#/contexts/StableSwapContext";
+
+export function ShareButton() {
+  const { generateURL } = useStableSwap();
+
+  const onInitialPage = usePathname() === "/stableswapsimulator";
+  const [isNotifierOpen, setIsNotifierOpen] = useState<boolean>(false);
+
+  return (
+    <ToastProvider>
```
luizakp, line 17: `<ToastProvider>` is not needed here as it is being invoked on the `<RootLayout>`

packages/next-app/src/app/stableswapsimulator/(components)/ShareButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+            <div className="flex h-14 flex-row items-center justify-between px-4 py-8">
+              <div className="flex flex-col justify-between space-y-1">
+                <h1 className="text-xl font-medium text-slate12">
+                  Link copied
+                </h1>
+                <h3 className="mb-2 text-sm leading-3 text-slate11">
+                  Use it to access the dashboard state.
+                </h3>
+              </div>
+            </div>
```
luizakp, line 56: Had to take a look at this 😅 but the toast was raising an error and the source was not having the content on its component.

This is because on toast we are cloning the props `content` and inserting a new prop called `close`. But your parent tag of content is a `div` and divs [don't accept custom props ](https://reactjs.org/link/attribute-behavior). That's why we have to create a component (local scoped, so inside this page with no export) to wrap up the divs and that component will receive the `close` prop

like this:

```js
      <Toast
        content={<ToastContent />}
        isOpen={isNotifierOpen}
        setIsOpen={setIsNotifierOpen}
        variant={"notification"}
      />
    </>
  );
}

function ToastContent() {
  return (
    <div className="flex h-14 flex-row items-center justify-between px-4 py-8">
      <div className="flex flex-col justify-between space-y-1">
        <h1 className="text-xl font-medium text-slate12">Link copied</h1>
        <h3 className="mb-2 text-sm leading-3 text-slate11">
          Use it to access the dashboard state.
        </h3>
      </div>
    </div>
  );
}
```

packages/next-app/src/contexts/StableSwapContext.tsx
```diff
@@ -53,7 +54,7 @@ export function StableSwapProvider({ children }: PropsWithChildren) {
   };
   const [initialData, setInitialData] =
     useState<AnalysisData>(defaultBaselineData);
-  const [customData, setCustomData] =
+  const [customData, setCustomData] = // rename function
```
luizakp, line 57:
```suggestion
  const [customData, setCustomData] =
```

luizakp (approved):
Great job 🥳

ribeirojose (approved):
🥳  only thing I'd add is that in your video we can't see the URLs 😝 but would send it to the Balancer folks as well!

PR #139, opened 2023-06-16
Handle desktop responsive resize on Metadata and Internal Manager

luizakp:
https://www.loom.com/share/7e3f7c8305f444faa45c9cc15497f50c?sid=11b2da00-e148-493d-be0a-473b1b0c4922


packages/next-app/src/app/stableswapsimulator/analysis/page.tsx
```diff
@@ -43,10 +43,10 @@ export default function Page() {
   }
 
   return (
-    <div className="flex flex-col w-full overflow-auto gap-y-20 pt-8 pr-8 max-h-[calc(100vh-132px)]">
+    <div className="flex max-h-[calc(100vh-132px)] w-full flex-col gap-y-20 overflow-auto pr-8 pt-8">
```
ribeirojose, line 46: I was testing something on my side and it seemed like `max-h-[calc(100vh-132px)]` was making the plots a bit weird on narrower screens (even a bit thinner than the 13" MBP). Have you tried removing it altogether?

packages/next-app/src/app/layout.tsx
```diff
@@ -13,7 +13,7 @@ export const metadata = {
 export default function Layout({ children }: { children: React.ReactNode }) {
   return (
     <html lang="en" className="h-full">
-      <body className="h-full flex flex-col bg-blue2">
+      <body className="flex h-full min-w-[930px] flex-col bg-blue2">
```
ribeirojose, line 16: on phones this would look super weird right?

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
```diff
@@ -91,25 +130,41 @@ function PoolCard({ isSelected, pool }: { isSelected: boolean; pool: Pool }) {
...
   const poolName =
     poolType === "Weighted" && tokens
-      ? tokens.map((obj) => obj.symbol).join("/")
+      ? tokens.map((obj) => obj.symbol).join(" / ")
       : name;
   const weights =
     poolType === "Weighted" && tokens
-      ? tokens.map((obj) => (Number(obj.weight) * 100).toFixed()).join("/")
+      ? tokens
+          .sort((a, b) => (a.weight < b.weight ? 1 : -1))
```
ribeirojose, line 138: What's the rationale behind sorting this? Does Balancer do it on their frontend? If not I'd suggest we keep to the same order returned form the subgraph

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
```diff
@@ -91,25 +130,41 @@ function PoolCard({ isSelected, pool }: { isSelected: boolean; pool: Pool }) {
...
+            "tracking-tighter text-lg font-bold text-slate12 max-h-7 truncate text-left",
+            weights ? "max-w-[60%] " : "w-full"
+          )}
+        >
+          {poolName}
+        </p>
+        {weights && (
+          <div className="max-w-[40%]">
+            <Badge isSelected={isSelected} isTrackingTighter>
+              {weights}
```
ribeirojose, line 157: I'd suggest to truncate this as well since you can always click & see the pool anyways

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
```diff
@@ -72,17 +93,35 @@ function OwnedPoolsSidebarItems({
...
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+          setPoolSearch(e.target.value)
+        }
+        value={poolSearch}
+      />
+      {filteredPools &&
+        (filteredPools.length === 0 ? (
+          <div className="flex flex-col text-slate12">
+            <span className="text-center">
+              Sorry, no pools were found &#128531;
```
ribeirojose, line 109: comment: `&#128531; = 😓`

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
```diff
@@ -72,17 +93,35 @@ function OwnedPoolsSidebarItems({
...
-            </Sidebar.Item>
-          </Link>
-        )
-      )}
-    </>
+    <div className="flex flex-col gap-y-2 pr-2">
+      <Input
+        label="Search pool"
+        placeholder="Search by name, symbol, id, tokens..."
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
```
ribeirojose, line 100: question: doesn't this get typed automatically?

ribeirojose (dismissed):
🙌 good stuff!

packages/next-app/src/app/stableswapsimulator/analysis/page.tsx
luizakp: yup, this PR is focused on the metadata and internal manager side of things. I trying to figure out the plots on stableswap. That's why decided to break into two PR. Sorry for not making that clear!

packages/next-app/src/app/layout.tsx
luizakp: From what I remembered on the beginning of Internal Manager is that our apps are not meant for phones rn. Are we going to tackle that?

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
luizakp: I think that if we truncate this information then its value is lost. I think we either show all the weights or we remove this badge

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
luizakp: This is automatically typed to `FormEvent<HTMLInputElement>` which provides no value on target 

```
Property 'value' does not exist on type 'EventTarget'.
```

packages/next-app/src/app/metadata/(components)/OwnedPoolsSidebarItems.tsx
ribeirojose: fwiw people can always go back to the pool page. But I'd lean towards having the same info that we have on the page to search the pool in the pool page.

luizakp: so we should add this pool's information here. Is that it?
image

PR #143, opened 2023-06-20
Jose/bal 425 dynamic format numbers based on the unit pt 2

ribeirojose:
- mostly refactors avoiding repeating code in plots + adding number formatting capability;
- adding the number formatting was simple, but I missed having some tests to make this refactor to ensure I wasn't breaking anything. I also think that if we had tests we this refactor wouldn't be needed because we'd already have noticed the redundancy/repetition and taken the chance to make things more consistent throughout the plots

### DIFF
--- a/packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
@@ -7,6 +7,59 @@ import { PlotType } from "plotly.js";
 import Plot, { defaultAxisLayout } from "#/components/Plot";
 import { Spinner } from "#/components/Spinner";
 import { AnalysisData, useStableSwap } from "#/contexts/StableSwapContext";
+import formatNumber from "#/utils/formatNumber";
+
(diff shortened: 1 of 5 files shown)


packages/next-app/src/app/stableswapsimulator/(components)/DepthCost.tsx
```diff
@@ -7,6 +7,59 @@ import { PlotType } from "plotly.js";
...
+    const displayAmount = `${formatNumber(amount, 2)} ${analysisSymbol}`;
+    const action =
+      direction === "in"
+        ? `${displayAmount} for ${tokenSymbols[i]}`
+        : `${tokenSymbols[i]} for ${displayAmount}`;
+
+    return `Swap ${action} to move the price ${tokenSymbols[i]}/${analysisSymbol} on ${templateDirection} <extra></extra>`;
+  });
+};
+
```
luizakp, line 29: 👏

luizakp (approved):
Using the DRY principle very well here! Thanks for putting attention into this matter and let's make does test 🦾

PR #146, opened 2023-06-20
[Stable Swap] Fix overflow x on analysis page pt.1

luizakp:
https://www.loom.com/share/b0a1e394b9e24efd8028fe6e42aac7c8?sid=d085d3f5-e462-4054-b5ca-269329b4a9a7

missing for pt2: sometimes the auto-resize from plotly does not work on shrink, from what I've tested the window resize event is not being always triggered


pnpm-lock.yaml
yvesfracari, line 1: Why are we changing this file without changing the `package.json`?

yvesfracari (dismissed):
Great work! I'll try to look into the plotly resize issue as well

pnpm-lock.yaml
ribeirojose: I think most of dependencies are locked to a range of versions in package.json, so running install from scratch (or maybe update? Not sure how pnpm treats this) can install them with the most up to date versions and therefore make changes to the lockfile. So probably some dependencies (or dependencies’ dependencies) got updated and our version range still matches (e.g. say we lock a dep to ^2.0.0, if a 2.1.x gets published it’s still within the version range of 2.0.0 (assuming you use semantic versioning) — more about that syntax here https://semver.npmjs.com/

packages/next-app/src/components/Table.tsx
```diff
@@ -37,16 +37,19 @@ export default function Table({
   color = "gray",
   variant = "solid",
   shade = "dark",
+  customClass,
```
ribeirojose, line 40: Would prefer the already established pattern “classNames”

ribeirojose (dismissed):
Realistic we don’t need to accommodate for the use case of someone resizing the page (although that’s another interesting problem!). I think that event might fire when changing screen orientation on mobile devices(tbc though) but also if someone resizes the window they can always refresh the page and that’s good enough for our use case imo. Wdyt? ive seen components behave weird after resizing even though if you load the page on that new viewport size they’ll behave normally 😅

luizakp:
@yvesfracari no need to look further into this, @ribeirojose and I talked that this is an interesting problem to solve but it doesn't bring much value to our application

PR #149, opened 2023-06-26
[Internal Manager] Remove gql query from homepage

luizakp:
This PR solves [this issue](https://linear.app/bleu-llc/issue/BAL-473/balancer-tools-network-handling-is-breaking)


packages/next-app/src/app/internalmanager/(components)/BatchWithdraeButton.tsx
```diff
@@ -0,0 +1,46 @@
+import { Address } from "@bleu-balancer-tools/shared";
```
ribeirojose, line 1: just noted a small typo in the component name 😄

PR #150, opened 2023-06-29
[Stable Swap] Balancer feedback

luizakp:
This PR solves the following feedback:
- Make the "Import pool parameters" larger/more obvious
- Load the app with a pool like wstETH/WETH already shown. It's the largest
- Can let users enter 4 for 4%  for the swap fee on frontend but calcs remain the same.

### DIFF
--- a/packages/next-app/src/app/internalmanager/(components)/BatchWithdrawButton.tsx
+++ b/packages/next-app/src/app/internalmanager/(components)/BatchWithdrawButton.tsx
@@ -8,7 +8,7 @@ import { getNetwork } from "#/contexts/networks";
 import { internalBalances } from "#/lib/gql";
 import { refetchRequest } from "#/utils/fetcher";
 
-export function WithdrawBatchButton() {
+export function BatchWithdrawButton() {
   const { address } = useAccount();
   const { chain } = useNetwork();
   const network = getNetwork(chain?.name);
--- a/packages/next-app/src/app/stableswapsimulator/(components)/CustomDataForm.tsx
+++ b/packages/next-app/src/app/stableswapsimulator/(components)/CustomDataForm.tsx
@@ -10,6 +10,7 @@ import { Select, SelectItem } from "#/components/Select";
 import { Form, FormField, FormLabel } from "#/components/ui/form";
 import { AnalysisData, useStableSwap } from "#/contexts/StableSwapContext";
 import useDebounce from "#/hooks/useDebounce";
+import { numberToPercent, percentToNumber } from "#/utils/formatNumber";
 
 import { TokenTable } from "./TokenTable";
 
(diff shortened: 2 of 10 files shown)


packages/next-app/src/app/internalmanager/(components)/BatchWithdrawButton.tsx
```diff
@@ -8,7 +8,7 @@ import { getNetwork } from "#/contexts/networks";
 import { internalBalances } from "#/lib/gql";
 import { refetchRequest } from "#/utils/fetcher";
 
-export function WithdrawBatchButton() {
```
yvesfracari, line 11: 👍

packages/next-app/src/app/stableswapsimulator/(components)/CustomDataForm.tsx
```diff
@@ -88,24 +95,28 @@ export default function CustomDataForm() {
...
+                }}
+                defaultValue={swapFeeInPercentage}
+                placeholder="Define the initial swap fee"
+              />
+            )}
+          />
+          <span className="absolute top-8 right-2 flex items-center text-slate10">
+            %
+          </span>
+        </div>
```
yvesfracari, line 119: Could check if we are accepting 100% of swap fee and display an error for this?

packages/next-app/src/app/stableswapsimulator/(components)/CustomDataForm.tsx
```diff
@@ -46,13 +47,19 @@ export default function CustomDataForm() {
     if (Object.keys(errors).length) return;
 
     const data = getValues();
-    setCustomData(data as AnalysisData);
+    const dataToCalculate = {
+      ...data,
+      swapFee: percentToNumber(data.swapFee),
+    };
+    setCustomData(dataToCalculate as AnalysisData);
```
yvesfracari, line 54: Think that would be easier to let this number as percent and transform it just on the math interface.

packages/next-app/src/utils/formatNumber.ts
```diff
@@ -17,5 +17,13 @@ const formatNumber = (
     style,
   });
 };
-
 export default formatNumber;
+
```
yvesfracari, line 21: Why we don't refactor the export default of the formatNumber?

yvesfracari (dismissed):
Just let some optional comments! Feel free to merge this PR!

packages/next-app/src/app/stableswapsimulator/(components)/CustomDataForm.tsx
luizakp: yeah, it was clear to me that this issue has different approaches, but only changing on math we would have still to transform the swapFee when importing a pool, as the data from to subgraph comes as the full number. 

Think our suggested approach is very valid, but at first, wanted to do both transformations in the same file

PR #152, opened 2023-07-10
[Internal Manager] Balancer's feedback

luizakp:
This PR solves:

- [You click on the links and sometimes it takes a while to load but UI is not communicating that with the user](https://linear.app/bleu-llc/issue/BAL-490/loader-on-link-route-push)
- [In Token Search might be better to make entire line selectable for existing tokens (currently have to click token name/icon or balance](https://linear.app/bleu-llc/issue/BAL-489/make-entire-row-of-token-list-clickable)
- [Goerli Chain Explorer linked only to connected Recipient instead of address entered](https://linear.app/bleu-llc/issue/BAL-491/goerli-chain-explorer-linked-only-to-connected-recipient-instead-of)

### DIFF
--- a/packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
+++ b/packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
@@ -10,7 +10,7 @@ import { BigNumber, formatFixed } from "@ethersproject/bignumber";
 import { ChevronDownIcon, MagnifyingGlassIcon } from "@radix-ui/react-icons";
 import { erc20ABI, fetchBalance, multicall } from "@wagmi/core";
 import Image from "next/image";
-import Link from "next/link";
+import { useRouter } from "next/navigation";
 import { tokenLogoUri } from "public/tokens/logoUri";
 import React, { useEffect, useState } from "react";
 import { useAccount, useNetwork } from "wagmi";
@@ -79,7 +79,6 @@ export function TokenSelect({
 }
 
 function TokenModal({
-  close,
   operationKind,
 }: {
   close?: () => void;
@@ -297,7 +296,6 @@ function TokenModal({
                     key={token.tokenAddress}
                     token={token}
                     operationKind={operationKind}
-                    close={close}
                     chainName={network}
                   />
                 );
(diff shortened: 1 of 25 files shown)


packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
```diff
@@ -312,85 +310,44 @@ function TokenModal({
...
+  const router = useRouter();
   return (
-    <Table.BodyRow key={token.tokenAddress} classNames="hover:bg-blue4">
+    <Table.BodyRow
+      key={token.tokenAddress}
+      classNames="hover:bg-blue4 hover:cursor-pointer"
+      onClick={() => {
+        router.push(
+          `/internalmanager/${chainName}/${operationKind}/token/${token.tokenAddress}`
+        );
```
ribeirojose, line 327: NextJS' docs recommends using the Link component when there are no strict requirements for using `router.push`: https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating. Did you find this performed better than using Link? It should literally be the same thing

packages/next-app/src/components/Table.tsx
```diff
@@ -98,9 +98,20 @@ function Body({ children }: React.PropsWithChildren) {
...
   classNames,
-}: React.PropsWithChildren<{ classNames?: string }>) {
+  onClick,
+}: React.PropsWithChildren<{ classNames?: string; onClick?: () => void }>) {
   useTableContext();
-  return <tr className={cn(classNames ?? classNames)}>{children}</tr>;
+  return (
+    <>
+      {onClick ? (
+        <tr className={cn(classNames ?? classNames)} onClick={onClick}>
```
ribeirojose, line 107: `classNames ?? classNames` shouldn't be needed since `cn` [can take an undefined string](https://github.com/JedWatson/classnames/blob/main/index.d.ts).

packages/next-app/src/components/Spinner.tsx
```diff
@@ -1,7 +1,45 @@
...
+          {
+            "h-4 w-4": SpinnerSizeNumber === SpinnerSize.sm,
+            "h-12 w-12": SpinnerSizeNumber === SpinnerSize.md,
+            "h-20 w-20": SpinnerSizeNumber === SpinnerSize.lg,
+          },
+          {
+            "border-amber10": SpinnerColorString === SpinnerColor.amber,
+            "border-blue10": SpinnerColorString === SpinnerColor.blue,
+          }
+        )}
```
ribeirojose, line 41: 👌

packages/next-app/src/components/Link.tsx
```diff
@@ -0,0 +1,30 @@
+"use client";
+
+import { Url } from "next/dist/shared/lib/router/router";
+import Link from "next/link";
+import React from "react";
+
+import { Spinner, SpinnerColor } from "./Spinner";
+
+export function LinkComponent({
```
ribeirojose, line 9: Cool! Would be helpful to see a demo of this to review

packages/next-app/src/components/Link.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const [isLoading, setIsLoading] = React.useState(false);
+
+  const ClonedElement = React.cloneElement(content, {
+    onClick: () => setIsLoading(true),
+    children: isLoading ? (
+      <Spinner color={loaderColor} size={"sm"} />
+    ) : (
+      content.props.children
+    ),
+  });
```
ribeirojose, line 27: Saw what you wanted to do here. You could also have [enforced a single child](https://react.dev/reference/react/Children#children-only) and skipped the added complexity here, wdyt? Am I missing something here?

ribeirojose (commented):
Sounds like most changes were changing `next/link:Link` calls to use the `LinkComponent`  you created, right?

packages/next-app/src/components/Link.tsx
luizakp: Sure! Now that the machine on fly is back to normal this will be probably less needed but still a great plan B to make sure the UX is great

https://www.loom.com/share/bec9ff4be2c74c61bb2334123414e9f1?sid=712e00da-6bbb-41bf-b645-a3a9a14efcaf

packages/next-app/src/components/Link.tsx
luizakp: not quite if I understood 😅

let's take this example: 

```js

    <LinkComponent
      href={`/internalmanager/${network}/withdraw/all`}
      content={
        <Button
          className="flex items-center gap-1"
          shade="light"
          variant="outline"
          title="Withdraw all"
          disabled={!hasTokenWithBalance}
        >
          <MinusIcon />
          Batch Withdraw
        </Button>
      }
    />
``` 

here I'm receiving `<Button>` and I want to keep everything the same but the inner text inside this button, which would be its child: `<MinusIcon /> Batch Withdraw`. So I'm using `cloneElement` to handle the `isLoading` state and change the inner Button text when true. From what I read in the documentation I cannot achieve the same with enforcing a single child. If it's possible can you give more details on that?

luizakp:
> Sounds like most changes were changing `next/link:Link` calls to use the `LinkComponent` you created, right?

@ribeirojose Yup, related to this feedback: [You click on the links and sometimes it takes a while to load but UI is not communicating that with the user](https://linear.app/bleu-llc/issue/BAL-490/loader-on-link-route-push)

packages/next-app/src/app/internalmanager/(components)/TokenSelect.tsx
ribeirojose: 😬 interesting! I think we could use TW's `table` classes to go around this, just FYI though. https://tailwindcss.com/docs/display#table

packages/next-app/src/components/Link.tsx
ribeirojose: That makes sense! I had in mind something that I tried implementing and landed in a very similar to what you had here 😝 your implementation is more direct though.

PR #153, opened 2023-07-10
[Internal Manager] Check allowance before asking to approve

luizakp:
https://linear.app/bleu-llc/issue/BAL-493/check-allowance-before-asking-to-approve
https://www.loom.com/share/f93700ad3e154948ab9698add5102cb7?sid=f6590b42-25d4-49fe-9010-1b9e209ed4e7

Todo:
- [ ] fetch all allowances on the first render and use it instead of reading the contract on every render
- [ ] clean the code/make more readable

### DIFF
--- a/packages/next-app/src/hooks/useTransaction.ts
+++ b/packages/next-app/src/hooks/useTransaction.ts
@@ -274,36 +274,44 @@ export function useInternalBalancesTransaction({
     args: [userBalancesOp],
   });
 
-  const { data, write } = useVaultManageUserBalance(config);
+  const { data, write, error } = useVaultManageUserBalance(config);
 
   //function to prepare depoist transaction
   async function approveToken() {
-    if (!submitData) return;
-    setTransactionStatus(TransactionStatus.WAITING_APPROVAL);
-    setNotification(
-      NOTIFICATION_MAP_INTERNAL_BALANCES[TransactionStatus.WAITING_APPROVAL]
-    );
-    const config = await prepareWriteContract({
-      address: submitData[0].tokenAddress as Address,
-      abi: erc20ABI,
-      functionName: "approve",
-      args: [
-        vaultAddress[5],
-        parseFixed(
-          submitData[0]?.tokenAmount ? submitData[0].tokenAmount : "0",
-          submitData[0]?.tokenDecimals ? submitData[0].tokenDecimals : "0"
-        ),
-      ],
-    });
-    const data = await writeContract(config);
-    const { hash, wait } = data;
-    handleTransactionStatus({ hash });
-    const receipt = await wait();
-    if (receipt.status) {
-      setTransactionStatus(TransactionStatus.CONFIRMING);
+    try {
+      if (!submitData) return;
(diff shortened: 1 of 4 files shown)


packages/next-app/src/hooks/useTransaction.ts
```diff
@@ -274,36 +275,75 @@ export function useInternalBalancesTransaction({
...
+      args: [userAddress, vaultAddress[5]],
     });
-    const data = await writeContract(config);
-    const { hash, wait } = data;
-    handleTransactionStatus({ hash });
-    const receipt = await wait();
-    if (receipt.status) {
+    const amountToApprove = parseFixed(tokenAmount, tokenDecimals);
+    if (allowance.gte(amountToApprove)) {
+      setHasAllowance(true); // User has enough allowance
```
mendesfabio, line 302: if you re-name `hasAllowance` to be more explicit maybe the comment isn't even needed, some examples: `isApprovalRequired` `hasEnoughAllowance` etc

mendesfabio (commented):
good progress, don't have much to comment beside the suggestion to only fetch allowance on load and maybe `multicall` over the token list.

packages/next-app/src/hooks/useTransaction.ts
```diff
@@ -343,6 +383,24 @@ export function useInternalBalancesTransaction({
...
+      setNotification(
+        NOTIFICATION_MAP_INTERNAL_BALANCES[TransactionStatus.WRITE_ERROR]
+      );
+      setTransactionStatus(TransactionStatus.CONFIRMING);
+    } else {
+      setNotification(
+        NOTIFICATION_MAP_INTERNAL_BALANCES[TransactionStatus.WRITE_ERROR]
+      );
+      setTransactionStatus(TransactionStatus.AUTHORIZING);
+    }
```
ribeirojose, line 401: Can you elaborate on why we need to set the transaction status after erroring out?

packages/next-app/src/hooks/useTransaction.ts
```diff
@@ -274,36 +275,75 @@ export function useInternalBalancesTransaction({
...
+        NOTIFICATION_MAP_INTERNAL_BALANCES[TransactionStatus.WAITING_APPROVAL]
       );
+      const config = await prepareWriteContract({
+        address: submitData[0].tokenAddress as Address,
+        abi: erc20ABI,
+        functionName: "approve",
+        args: [
+          vaultAddress[5],
+          parseFixed(
+            submitData[0]?.tokenAmount ? submitData[0].tokenAmount : "0",
```
ribeirojose, line 326: wouldn't `submitData[0]?.tokenAmount || "0",` work?

packages/next-app/src/hooks/useTransaction.ts
luizakp: So the user can reject on metamask and still be able to try again https://www.loom.com/share/32d8aad314204ed3922d10062f4ba639?sid=46aefd58-a45b-4944-af3d-41e26e516747

PR #156, opened 2023-07-13
refactor math package

yvesfracari:
Refactor math package to be more generic. The idea is to design the AMM class to be generic and receive multiple kinds of Pools (or math implementations). Also, some parts of our code were replaced for Balancer-sor implementation (such as the rates part).

https://excalidraw.com/#room=7c0c310605770ecdabf0,SVjmhsLQJC448N2wejB5DQ

This PR:
- Create refactored math package as `math-new`;
- Solve TS errors on Stable Swap Simulator
- Create Pool Simulator page (and context) integrated with the new package


packages/math-new/classes.svg
luizakp, line 1: maybe we should have a readme for the math to display this

packages/math-new/src/index.ts
```diff
@@ -0,0 +1,239 @@
...
+    const amountIn = this.math._tokenInForExactTokenOut(
+      poolPairData,
+      bnum(amountOut)
+    );
+    const effectivePrice = bnum(amountOut).div(amountIn);
+    const spotPrice = this.math._spotPrice(poolPairData);
+    return bnum(1).minus(effectivePrice.div(spotPrice)).toNumber();
+  }
+
+  priceImpactForExactTokenOutReversedSwap(
```
luizakp, line 225: all the AMM should have this function as a common point or this is just the example?

luizakp (changes requested):
Great job, @yvesfracari !

We just missed something on the math side during the refactor. I just cross-checked the same from production and from your branch and I got different results (maybe my branch is not entirely up to date due to the rebase to rewrite the commit history):

Production:
image

image


Localhost:
image

image

ribeirojose (dismissed):
a bit hard to review in detail, but looking good overall

PR #161, opened 2023-07-18
Update balancer tools packages

ribeirojose:


packages/next-app/package.json
```diff
@@ -16,12 +16,13 @@
...
   },
   "dependencies": {
-    "@balancer-labs/sor": "4.1.1-beta.9",
+    "@balancer-labs/sor": "4.1.1-beta.13",
     "@bleu-balancer-tools/gql": "workspace:*",
     "@bleu-balancer-tools/math": "workspace:*",
     "@bleu-balancer-tools/schema": "workspace:*",
     "@bleu-balancer-tools/tsconfig": "workspace:*",
-    "@hookform/resolvers": "^3.1.0",
+    "@ethersproject/constants": "^5.7.0",
```
luizakp, line 24: I think we can remove this one now

packages/utils/src/parseFixed.test.ts
luizakp, line 1: 👏

PR #163, opened 2023-07-19
fix: inverse price base to match SOR pattern

yvesfracari:
Checking the error that @luizakp pointed at the last PR I noticed that we used a different price definition. In SOR, the price definition is `tokenIn`/`tokenOut`. So, this PR uses this definition in the new math package and new UI.

obs:
- README about how to add new pool math missing


packages/math-new/README.md
```diff
@@ -0,0 +1,11 @@
+# Overview
+The Balancer SOR Economics Extension is a package designed to extend the functionality of the Balancer SOR package. This extension introduces essential economics concepts, such as price impact and effective price calculations, into the package.
+
+# Package Structure
+The package is structured around a main Automated Market Maker (AMM) class that serves as the central point of interaction for users. The main components of the package are:
+
+- AMM Class: The AMM class incorporates all the economic functions related to the Balancer SOR pools. It provides methods to calculate price impact, effective price, and other economic indicators for trades.
+
+- SOR Extended Pool Classes: The package uses the Balancer SOR Pool classes as attributes in the AMM class. These classes from the Balancer SOR pools package handle the core functionality of interacting with Balancer SOR pools. Usually, we need to add a few functions on the SOR package (as an spot price function). Those functions should use Balancer SOR code as reference or even copied.
```
ribeirojose, line 9: Make sure we're referencing them back when copied and to mention that we've done that because they're not exported in the main package. I'd also consider forking the SOR repo ourselves and introducing some of these changes there -- at least the ones regarding exporting of functions (like @luizakp did [here](https://github.com/bleu-studio/balancer-sor/pull/1)) -- and making a PR back to their main repo, so that we don't have to worry about changes to their code not being reflected back on our side. WDYT?

Would also take a look into https://github.com/balancer/b-sdk which seems to be the unification of the SOR with the SDK and peek into what they're set out to do there -- I'd be curious about whether could help that effort somehow.

packages/math-new/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Note: In the context of this package, symbols are used to represent tokens, allowing both real tokens with addresses and imaginary tokens used in the dashboard environment to be handled seamlessly. This is particularly useful when dealing with the dashboard that supports imaginary tokens. Each extended pool class must handle this.
```
ribeirojose, line 11: "Each extended pool class must handle this." what's this referring to here?

packages/math-new/src/index.ts
```diff
@@ -41,19 +41,19 @@ export class AMM {
     switch (amm.poolType) {
       case "MetaStable":
         this.math = new ExtendedMetaStableMath(
-          amm.poolParams as IMetaStableMath,
+          amm.poolParams as IMetaStableMath
         );
       case "MetaStable_2":
```
ribeirojose, line 46: `#comment` "MetaStable_2" is just to serve as a placeholder so that Lu can start working on the Gyro stuff.

packages/math-new/src/index.ts
```diff
@@ -41,19 +41,19 @@ export class AMM {
...
+          amm.poolParams as IMetaStableMath_2
         );
     }
   }
 
   exactTokenInForTokenOut(
     amountIn: number,
     tokenIn: string,
-    tokenOut: string,
+    tokenOut: string
```
ribeirojose, line 56: think this is probably a prettier change 🤔 have you run prettier in this PR after merging back from the target branch?

ribeirojose (approved):
🙌 looking good

PR #165, opened 2023-07-21
[Pool Simulator] Load Pool from MetaStable and GyroE and reset pool params when pool type changes

yvesfracari:
This PR:
 -  [Fix load existing pool](https://linear.app/bleu-llc/issue/BAL-528/fix-load-existing-pool)
 - [Fill / Reset pool parameters when change pool type](https://linear.app/bleu-llc/issue/BAL-529/fill-reset-pool-parameters-when-change-pool-type)

https://www.loom.com/share/1ef8292e15854631bceb5341b54c844b?sid=a23e670f-0989-4db5-b196-c9bf616ae13e


devjoaov (commented):
@luizakp From what I understood here you are using the poolSimulator context to define the initial pool type. In addition, there is a select to change the pool type, resulting in different fields for the form. There is also a confirmation screen since the user can lose the parameters that he previously configured, right?

I think that here would be a good case for using the xstate package itself, in it you can have an initial state (you can define which pool type you want), and a context pooltype1 that is an object with the parameters of pooltype1 and a poolType2 that is another object with the parameters for poolType2 (check https://xstate.js.org/docs/guides/context.html#initial-context)

You can create two states that would be the states of pool types, and when changing states you update the current context (it will be like saving the parameters that the user selected, so you don't need the confirmation modal since it wouldn't lose the parameters). For each state, you can render a different form (or component)

apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
```diff
@@ -6,11 +6,12 @@ import { PlotType } from "plotly.js";
...
 import Plot, { defaultAxisLayout } from "#/components/Plot";
 import { Spinner } from "#/components/Spinner";
-import { TokensData, useStableSwap } from "#/contexts/PoolSimulatorContext";
+import { TokensData, usePoolSimulator } from "#/contexts/PoolSimulatorContext";
 import { formatNumber } from "#/utils/formatNumber";
 
 export function DepthCost() {
-  const { analysisToken, initialData, initialAMM, customAMM } = useStableSwap();
+  const { analysisToken, initialData, initialAMM, customAMM } =
+    usePoolSimulator();
```
yvesfracari, line 14: 👍

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -26,33 +39,113 @@ function AnalysisMenu() {
...
+            </div>
+          </Dialog>
+          <PoolParamsForm />
+        </div>
+      </div>
     </SearchPoolFormWithDataForm>
   );
 }
 
+function SelectPoolType({
```
yvesfracari, line 110: Maybe we could separate this component into a new file

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -26,33 +39,113 @@ function AnalysisMenu() {
...
+    </div>
+  );
+}
+
 function SearchPoolFormWithDataForm({ children }: { children: ReactElement }) {
+  const { poolType } = usePoolSimulator();
   return (
     <div>
-      <SearchPoolFormDialog>
+      <SearchPoolFormDialog poolTypeFilter={poolType}>
```
yvesfracari, line 148: 👍

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+  StableSwapSimulatorDataSchema,
+} from "#/lib/schema";
+
+const schemaMapper = {
+  MetaStable: StableSwapSimulatorDataSchema,
+  GyroE: ECLPSimulatorDataSchema,
+};
+
+interface IInput {
+  name: "swapFee" | "ampFactor" | "alpha" | "beta" | "lambda" | "c" | "s";
```
yvesfracari, line 25: Isn't there a way to get this from all `poolParams` type?

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+                  {...field}
+                  label={input.label}
+                  type="number"
+                  validation={{
+                    required: true,
+                    valueAsNumber: true,
+                    value: initialData.poolParams?.[input.name],
+                  }}
+                  defaultValue={initialData.poolParams?.[input.name]}
+                  placeholder="Define the initial value "
```
yvesfracari, line 163:
```suggestion
                  placeholder={input.placeholder}
```

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -0,0 +1,137 @@
+import { PoolQuery } from "@bleu-balancer-tools/gql/src/balancer/__generated__/Ethereum";
```
yvesfracari, line 1: 👍

apps/balancer-tools/src/components/Dialog.tsx
```diff
@@ -12,30 +12,39 @@ export function Dialog({
...
+  isOpen?: boolean;
+  setIsOpen?: (isOpen: boolean) => void;
+  onClose?: (event: Event) => void;
 }>) {
   const [open, setOpen] = React.useState(false);
+  const isDialogOpen = isOpen ?? open;
+  const isDialogSetOpen = setIsOpen ?? setOpen;
   return (
-    <DialogPrimitive.Root open={open} onOpenChange={setOpen}>
+    <DialogPrimitive.Root open={isDialogOpen} onOpenChange={isDialogSetOpen}>
```
yvesfracari, line 32: 🤔 Didn't completely get why we need this logic. Could you elaborate?

yvesfracari (commented):
I can't approve bc I did some of the commits but it is good to go for me. Just double-check if the schema changes don't break the stableswap project. All other comments are optional.

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
luizakp: Yes! I did this on https://github.com/bleu-studio/balancer-tools/blob/6c00db3d82cdfc1a41bb49c9df9c789283a1847c/apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx#L23

apps/balancer-tools/src/components/Dialog.tsx
luizakp: Sure! By using const `[open, setOpen] = React.useState(false);` the dialog would be open on the `onClick` of the child component. But in the case of a Select, we want the dialog to be open only when the Select value changes. To achieve this I used an external state, but it's optional because if you don't provide the state the Dialog will open on `onClick` of the child

luizakp: so basically now we have two options to trigger the Dialog:
- Default: `onClick` of child component
- Custom: handle the open state outside of Dialog and use inside of it

Let me know if this was confusing 😅

apps/balancer-tools/src/lib/schema.ts
```diff
@@ -173,15 +173,30 @@ export const getStableSwapSimulatorTokensSchema = ({
   return StableSwapSimulatorTokensSchema;
 };
 
-export const StableSwapTokensSchema = z.object({
+export const TokensSchema = z.object({
```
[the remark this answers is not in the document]
luizakp: Almost sure this is not being used on `stableswap`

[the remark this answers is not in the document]
luizakp: This is only used on `schema.ts`

```
export const StableSwapTokensSchema = z.object({
  symbol: z.string().min(1),
  balance: z.coerce.number().positive(),
  rate: z.coerce.number().positive(),
  decimal: z.coerce.number().int().positive().max(60),
});

export const StableSwapSimulatorDataSchema = z.object({
  swapFee: z.coerce.number().positive().min(0.0001).max(10), //source: https://github.com/balancer/balancer-v2-monorepo/blob/c4cc3d466eaa3c1e5fa62d303208c6c4a10db48a/pkg/pool-utils/contracts/BasePool.sol#L74
  ampFactor: z.coerce.number().positive().min(1).max(5000), //source: https://github.com/balancer/balancer-v2-monorepo/blob/c4cc3d466eaa3c1e5fa62d303208c6c4a10db48a/pkg/pool-stable/contracts/StableMath.sol#L28
  tokens: z.array(StableSwapTokensSchema).min(2),
});
```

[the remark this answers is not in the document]
luizakp: Then I changed on `StableSwapSimulatorDataSchema` https://github.com/bleu-studio/balancer-tools/blob/009b32763cb0fbd4332500c9ca8ac8daa86af6b4/apps/balancer-tools/src/lib/schema.ts#L201

apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -16,23 +12,62 @@ import {
...
 
-export interface AnalysisData {
-  tokens: TokensData[];
+export interface MetaStableParams {
   ampFactor?: number;
   swapFee?: number;
 }
 
-interface StableSwapContextType {
+export interface GyroEParams {
```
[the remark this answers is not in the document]
luizakp: fixed this on https://github.com/bleu-studio/balancer-tools/pull/171/commits/c6a2cf0963f012b285b74c7babfe0fb461a05449

PR #166, opened 2023-07-23
[Internal Manager] Add polygon , fix Optimism and remove Sepolia from networks

luizakp:
This PR:
- adds Polygon
- fix Optimism url check on `getNetwork`
- remove sepolia from networks, because viem is not recognizing successful transactions
- remove `Subgraph.BalancerInternalManager` since `Subgraph.BalancerPools` now has `tokenInfo` on `userInternalBalances` query

Tested networks - Show token table:
- Ethereum -> Ok;
- Goerli -> Ok;
- Sepolia -> Ok;
- Polygon -> Ok;
- Polygon zkevm -> Ok;
- Gnosis -> Ok;
- Arbitrum -> Ok;
- Optimism -> Ok;

Tested networks - Transactions:
- Ethereum -> Ok;
- Goerli -> Ok;
- Sepolia -> Failed. The transaction occurs successfully, but viem doesn't find the transaction hash, invoking an error toast, when actually the transaction goes fine

### DIFF
--- a/apps/balancer-tools/src/app/internalmanager/layout.tsx
+++ b/apps/balancer-tools/src/app/internalmanager/layout.tsx
@@ -29,6 +29,8 @@ export default function Layout({ children }: { children: React.ReactNode }) {
               Network.Ethereum,
               Network.Gnosis,
               Network.Arbitrum,
+              Network.Polygon,
+              Network.PolygonZKEVM,
               Network.Optimism,
               Network.Goerli,
               Network.Sepolia,
--- a/apps/balancer-tools/src/app/internalmanager/layout.tsx
+++ b/apps/balancer-tools/src/app/internalmanager/layout.tsx
@@ -33,7 +33,6 @@ export default function Layout({ children }: { children: React.ReactNode }) {
               Network.PolygonZKEVM,
               Network.Optimism,
               Network.Goerli,
-              Network.Sepolia,
             ]}
             chainName={network}
           >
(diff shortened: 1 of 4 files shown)


ribeirojose:
> This PR:
> ...
> * Sepolia -> Failed. The transaction occurs successfully, but viem doesn't find the transaction hash, invoking an error toast, when actually the transaction goes fine

I think we haven't seen this pattern in other apps right? Think we should log it to explore later?

luizakp:
> > This PR:
> > ...
> > 
> > * Sepolia -> Failed. The transaction occurs successfully, but viem doesn't find the transaction hash, invoking an error toast, when actually the transaction goes fine
> 
> I think we haven't seen this pattern in other apps right? Think we should log it to explore later?

@ribeirojose 

Totally, it seemed something between viem and sepolia rather than something that we did because mainnet and goerli worked as expected. And as Sepolias is a testnet we don't need to rush this IMO

ribeirojose:
> > > This PR:
> > > ...
> > > 
> > > * Sepolia -> Failed. The transaction occurs successfully, but viem doesn't find the transaction hash, invoking an error toast, when actually the transaction goes fine
> > 
> > 
> > I think we haven't seen this pattern in other apps right? Think we should log it to explore later?
> 
> @ribeirojose
> 
> Totally, it seemed something between viem and sepolia rather than something that we did because mainnet and goerli worked as expected. And as Sepolias is a testnet we don't need to rush this IMO

Also wdyt @mendesfabio? Could it be related to Sepolia's deployment? Is Balancer's deployment complete there?

mendesfabio:
Vault has been deployed and that should be enough. I'd say this is probably related to RPC providers, let me try to reproduce it.

PR #168, opened 2023-07-24
use interface IAMMFunctionality to abstract math pkgs common interface

ribeirojose:


packages/math-poolsimulator/src/index.ts
```diff
@@ -26,266 +27,166 @@ export function numberToBigNumber({
...
-      const poolPairData = this.math.parsePoolPairData(
-        tokenIn,
-        tokenOut
-      ) as GyroEPoolPairData;
+      );
       return this.math._spotPrice(poolPairData).toNumber();
-    } else throw new Error("Invalid math instance");
   }
 
-  _tokenInForExactSpotPriceAfterSwap({
```
ribeirojose, line 124: This one is commented out because it really changes depending on the pool type right?

PR #170, opened 2023-07-27
Add tokens table on Pool Params Form pt.1

yvesfracari:
This PR adds the tokens table fully integrated with the pool parameter component but without considering the second step (this will be pt 2 of the issue)


apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -109,27 +113,10 @@ export function PoolParamsForm() {
...
-            z: initialData.poolParams?.z,
-          }
-        : {};
-
     const dataWithPoolType = {
-      poolParams: {
-        ...data,
-        ...additionalParams,
-      },
+      poolParams: Object.fromEntries(
```
luizakp, line 117: 👏

apps/balancer-tools/src/app/poolsimulator/analysis/page.tsx
```diff
@@ -21,6 +21,24 @@ export default function Page() {
...
+              <span className="text-slate12">index</span>
+              <span className="text-slate12">{index}</span>
+              <span className="text-slate12">Symbol</span>
+              <span className="text-slate12">{token.symbol}</span>
+              <span className="text-slate12">Balance</span>
+              <span className="text-slate12">{token.balance}</span>
+              <span className="text-slate12">Rate</span>
+              <span className="text-slate12">{token.rate}</span>
+              <span className="text-slate12">Weight</span>
+              <span className="text-slate12">{token.weight}</span>
```
luizakp, line 37: I think we are never requesting the token weight

```
export const poolById = gql`
  query Pool($poolId: ID!) {
    pool(id: $poolId) {
      address
      owner
      poolType
      symbol
      swapFee
      amp
      c
      s
      alpha
      beta
      lambda
      tauAlphaX
      tauAlphaY
      tauBetaX
      tauBetaY
      u
      v
      w
      z
      dSq
      tokens {
        address
        symbol
        balance
        decimals
        priceRate
      }
    }
  }
`;
```

apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -118,24 +119,25 @@ export function PoolSimulatorProvider({ children }: PropsWithChildren) {
...
+  // }, [initialData]);
+
+  // useEffect(() => {
+  //   if (
+  //     !customData.poolType &&
+  //     !customData.poolParams?.swapFee // all pool type have swapFee
+  //   )
+  //     return;
+  //   setCustomAMM(convertAnalysisDataToAMM(customData));
+  // }, [customData]);
```
luizakp, line 140: IMO we should delete the code we are not using at the moment because we still have this as a reference either on stableswap or in our git commit history. WDYT?

luizakp (approved):

apps/balancer-tools/src/app/poolsimulator/analysis/page.tsx
yvesfracari: For now yes (but I think that I should remove this weights for now)

PR #172, opened 2023-07-28
[Pool Simulator] Implement form second step

yvesfracari:
https://www.loom.com/share/3375b9a8d81743e0a66737056ea6ca56

Closes BAL-514
Closes BAL-499


apps/balancer-tools/src/contexts/FormContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { AnalysisData, usePoolSimulator } from "./PoolSimulatorContext";
+
+interface PoolSimulatorContextType {
+  data: AnalysisData;
+  setData: (data: AnalysisData) => void;
+  isCustomData: boolean;
+}
+
+export const FormContext = createContext({} as PoolSimulatorContextType);
```
ribeirojose, line 13: `#must` rename this not to be mistaken by react-hook-form's own form context

apps/balancer-tools/src/components/ui/form.tsx
```diff
@@ -35,7 +35,10 @@ const Form = <T extends FieldValues>({
   return (
     <FormProvider {...formMethods}>
       <form
-        onSubmit={onSubmit}
+        onSubmit={(e) => {
+          e.stopPropagation();
+          onSubmit(e);
+        }}
```
ribeirojose, line 41: 👍

apps/balancer-tools/src/app/poolsimulator/analysis/page.tsx
```diff
@@ -3,40 +3,46 @@
...
+                  <div className="flex items-center gap-x-4">
+                    <span className="text-slate12">index</span>
+                    <span className="text-slate12">{index}</span>
+                    <span className="text-slate12">Symbol</span>
+                    <span className="text-slate12">{token.symbol}</span>
+                    <span className="text-slate12">Balance</span>
+                    <span className="text-slate12">{token.balance}</span>
+                    <span className="text-slate12">Rate</span>
+                    <span className="text-slate12">{token.rate}</span>
+                  </div>
```
ribeirojose, line 39:
```suggestion
                  <div className="flex items-center gap-x-4 text-slate12">
                    <span>index</span>
                    <span>{index}</span>
                    <span>Symbol</span>
                    <span>{token.symbol}</span>
                    <span>Balance</span>
                    <span>{token.balance}</span>
                    <span>Rate</span>
                    <span>{token.rate}</span>
                  </div>
```

apps/balancer-tools/src/app/poolsimulator/analysis/page.tsx
```diff
@@ -3,40 +3,46 @@
...
-              <span className="text-slate12">{token.symbol}</span>
-              <span className="text-slate12">Balance</span>
-              <span className="text-slate12">{token.balance}</span>
-              <span className="text-slate12">Rate</span>
-              <span className="text-slate12">{token.rate}</span>
+    <div>
+      {[initialData, customData].map((data) => {
+        return (
+          <div className="flex lg:max-h-[calc(100vh-132px)] w-full flex-col gap-y-20 lg:overflow-auto pr-8 pt-8">
+            {/* (h-screen - (header's height + footer's height)) = graph's height space */}
```
ribeirojose, line 13: what's this comment about? 👀

apps/balancer-tools/src/app/poolsimulator/(components)/SelectPoolType.tsx
```diff
@@ -1,17 +1,19 @@
...
-export function SelectPoolType({
-  selectedType,
-}: {
-  selectedType: PoolType;
-  setSelectedType: (value: PoolType) => void;
-}) {
-  const { setPoolType } = usePoolSimulator();
+export function SelectPoolType({ selectedType }: { selectedType: PoolType }) {
+  const { setData, data } = usePoolFormContext();
+  const event = new CustomEvent("changePoolType");
```
ribeirojose, line 8: can you elaborate more about why you chose to create a custom event here?

apps/balancer-tools/src/app/poolsimulator/(components)/SearchPoolFormDialog.tsx
```diff
@@ -17,14 +18,22 @@ export function SearchPoolFormDialog({
   poolTypeFilter,
 }: PropsWithChildren<{ poolTypeFilter: PoolType }>) {
   const { handleImportPoolParametersById } = usePoolSimulator();
+  const { setData, data, isCustomData } = usePoolFormContext();
+  const onSubmit = (formData: PoolAttribute) => {
+    if (isCustomData) {
+      handleImportPoolParametersById(formData, setData, false, data);
+      return;
+    }
+    handleImportPoolParametersById(formData, setData);
```
ribeirojose, line 27: Couldn't it be
```suggestion
    handleImportPoolParametersById(formData, setData, !isCustomData, data);
```

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -28,6 +25,8 @@ interface IInput {
   label: string;
   placeholder: string;
   unit: string;
+  transformFromDataToForm: (n: number | undefined) => number | undefined;
+  transformFromFormToData: (n: number | undefined) => number | undefined;
```
ribeirojose, line 29:
```suggestion
  transformDataToForm: (n?: number) => number | undefined;
  transformFormToData: (n?: number) => number | undefined;
```

apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
```diff
@@ -21,7 +21,7 @@ export function ShareButton() {
             : "Copy a link with the current simulation parameters"
         }
       >
-        <Button shade="dark" disabled={onInitialPage}>
+        <Button shade="dark" disabled={true}>
```
ribeirojose, line 24: Do we need to come back to this later? If so I'd suggest we add a TODO with a task reference here.

apps/balancer-tools/src/contexts/FormContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+        setData: setInitialData,
+        isCustomData: false,
+      }}
+    >
+      {children}
+    </FormContext.Provider>
+  );
+}
+
+export function CustomFormContextProvider({ children }: PropsWithChildren) {
```
ribeirojose, line 31: can you say more as to what made you decide to use two contexts here instead of one?

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -26,8 +31,82 @@ const POOL_TYPES_MAPPER = {
 };
 
 function IndexMenu() {
+  const { push } = useRouter();
+  const { setCustomData, customData, setIsGraphLoading } = usePoolSimulator();
+  const [tabValue, setTabValue] = useState("initialData");
+  const clickInitialDataTab = new CustomEvent("clickInitialDataTab");
+  const clickCustomDataTab = new CustomEvent("clickCustomDataTab");
```
ribeirojose, line 38: same here, can you elaborate on the choice to use custom events here?

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -26,8 +31,82 @@ const POOL_TYPES_MAPPER = {
...
+            }}
+          >
+            <span>Custom</span>
+          </Tabs.ItemTrigger>
+        </Tabs.ItemTriggerWrapper>
+        <Tabs.ItemContent tabName="initialData">
+          <InitialFormContextProvider>
+            <SearchPoolFormWithDataForm>
+              <div className="flex flex-col mt-4">
+                <FormWithPoolType extraOnSubmit={initialFormExtraOnSubmit} />
```
ribeirojose, line 82:
```suggestion
                <FormWithPoolType onSubmit={initialFormExtraOnSubmit} />
```

apps/balancer-tools/src/app/poolsimulator/(components)/TokenTable.tsx
```diff
@@ -24,83 +24,53 @@ const MAX_POOL_TOKENS_MAPPER = {
...
+          <Table.HeaderCell padding={customPadding}>
+            <span className="sr-only">Edit</span>
+          </Table.HeaderCell>
+          <Table.HeaderCell padding={customPadding}>Symbol</Table.HeaderCell>
+          <Table.HeaderCell padding={customPadding}>Balance</Table.HeaderCell>
+          <Table.HeaderCell padding={customPadding}>Rate</Table.HeaderCell>
+          <Table.HeaderCell padding={customPadding}>
+            {aboveOrEqualLimit && (
+              <Tooltip
+                content={`This pools type can't have more than ${MAX_POOL_TOKENS} tokens.`}
```
ribeirojose, line 45:
```suggestion
                content={`This pool type can't have more than ${MAX_POOL_TOKENS} tokens.`}
```

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -42,42 +121,34 @@ function IndexMenu() {
...
-                    {POOL_TYPES_MAPPER[type]}
-                  </SelectItem>
-                ))}
-              </Select>
-            </div>
-          </Dialog>
-          <PoolParamsForm />
+    <div className="flex flex-col">
+      <Dialog
+        content={<SelectPoolType selectedType={selectedType} />}
```
ribeirojose, line 126:
```suggestion
        content={<PoolTypeChangeConfirmation selectedType={selectedType} />}
```

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
-    getValues,
     clearErrors,
+    watch,
     formState: { errors },
   } = form;
 
-  const onSubmit = (data: FieldValues) => {
-    setIsGraphLoading(true);
+  const getOnSubmit = (tabClicked: boolean) => (fieldData: FieldValues) => {
+    const hasNullData = inputMapper[data.poolType].reduce(
```
ribeirojose, line 127:
```suggestion
    const hasNullData = poolTypeInputParamsMap[data.poolType].reduce(
```

ribeirojose: wouldn't this be already covered by validations?

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
 
+  function checkDataIsEqualForm() {
+    const fieldData = watch();
+    const dataTransformed = {
+      poolParams: Object.fromEntries(
+        inputMapper[data.poolType].map((input) => [
+          input.name,
+          input.transformFromDataToForm(fieldData[input.name]),
+        ])
+      ),
```
ribeirojose, line 154: could refactor this into its own function since it's also used below

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
     };
-
-    setInitialData(dataWithPoolType as AnalysisData);
-    setCustomData(dataWithPoolType as AnalysisData);
-    push("/poolsimulator/analysis");
+    setData(dataWithPoolType as AnalysisData);
+    extraOnSubmit?.(dataWithPoolType as AnalysisData, tabClicked);
   };
 
+  function checkDataIsEqualForm() {
```
ribeirojose, line 146: curious about what we were trying to achieve with this function

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
+        });
       });
-    });
-    setInitialData({
-      poolParams: undefined,
-      tokens: [],
-      poolType: poolType,
-    });
-  }, [poolType]);
+    }
```
ribeirojose, line 181: there's a `reset` method in react-hook-form to clear values/errors. Is that what you're trying to achieve here?

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
     clearErrors,
+    watch,
     formState: { errors },
   } = form;
 
-  const onSubmit = (data: FieldValues) => {
-    setIsGraphLoading(true);
+  const getOnSubmit = (tabClicked: boolean) => (fieldData: FieldValues) => {
+    const hasNullData = inputMapper[data.poolType].reduce(
+      (sum, input) => sum || !fieldData[input.name],
```
ribeirojose, line 128: you could also use `.some(fn)` to achieve the same result here

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
+      poolParams: Object.fromEntries(
+        inputMapper[data.poolType].map((input) => [
+          input.name,
+          input.transformFromDataToForm(fieldData[input.name]),
+        ])
+      ),
+      tokens: fieldData.tokens,
+      poolType: data.poolType,
+    };
+    return JSON.stringify(dataTransformed) == JSON.stringify(data);
```
ribeirojose, line 158: I don't think JSON.stringify guarantees the order of keys is the same, but I think in your scenario you only want to ensure that the two objects have the same data, right?

ribeirojose (changes requested):
In the right direction, but there's a couple sharp edges that need to be polished and I'm curious about . Don't think we should merge unless there's a very clear rationale behind that, since it's non standard, especially in a [React-ive app](https://react.dev/learn/lifecycle-of-reactive-effects).

apps/balancer-tools/src/app/poolsimulator/analysis/page.tsx
yvesfracari: this comment came from the stable swap. It is related with the max-h calculation

apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
yvesfracari: Ow, I think that was a problem dealing with conflicts. Luiza had change this file and change It back later.

apps/balancer-tools/src/app/poolsimulator/(components)/SelectPoolType.tsx
yvesfracari: This is needed to reset the form when the user changes the poolType.

The other solution I tried was to use an `useEffect` on the `data.poolType` on the forms component. However, when we changed from initial to custom data this `useEffect` was being triggered leading to form resetting when the user navigated with the tabs.

So this event is to be watched on the form component to reset the fields. I also thought about using a flag state to manage this, but the event seems more appropriate. What is your opinion about it?

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
yvesfracari: I just want to make clear that we already have another `onSubmit` on the `FormWithPoolType` component and this function will just extend the existing function.

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
-      poolType,
+      tokens: fieldData.tokens,
+      poolType: data.poolType,
     };
-
-    setInitialData(dataWithPoolType as AnalysisData);
-    setCustomData(dataWithPoolType as AnalysisData);
-    push("/poolsimulator/analysis");
+    setData(dataWithPoolType as AnalysisData);
+    extraOnSubmit?.(dataWithPoolType as AnalysisData, tabClicked);
```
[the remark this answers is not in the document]
yvesfracari: Yes! In my view, the standard behavior of this on submit is to save the data that was being modified. However, we can have to do other things depending on the context that the form is being used (such as changing the tab/page, setting the other data...). This would also be useful on the `analysis` page where we would probably have to just save the data.

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,110 +58,155 @@ const inputMapper: InputMapperType = {
...
     setValue,
-    getValues,
     clearErrors,
+    watch,
     formState: { errors },
   } = form;
 
-  const onSubmit = (data: FieldValues) => {
-    setIsGraphLoading(true);
+  const getOnSubmit = (tabClicked: boolean) => (fieldData: FieldValues) => {
```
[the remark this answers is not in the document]
yvesfracari: This is used to cover the cases on the `customForm`. If the user clicks on the `initialTab` we just want to save the changes of the `customData`, but if he clicks on the next step button we want also to push to the analysis page.

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
yvesfracari: We used to have this bc of a forever loop on the `useEffect` bellow. I tested now and it is not needed anymore (maybe bc of the changes that we made on the forms -- not sure). So I will remove this function.

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
luizakp: it's because we want to reset the inputs to `undefined` to show the placeholder for empty data. But `reset()` expects default data, and with `undefined` the form holds to the last defined value. 

If our default data was like swapFee: 0, then `reset()` works but it won't show the input placeholder

ribeirojose:
copilot:poem

ribeirojose (approved):
we worked this morning on some refactors in #174. Looking sharp 🔪

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -55,149 +45,183 @@ const inputMapper: InputMapperType = {
...
+    const { value: currentTab } = useTabContext();
+
+    useEffect(() => {
+      const data = getValues();
+      onTabChanged(createPayload(poolType, data));
+    }, [currentTab]);
+
+    // useEffect(()=> {
+    //   if (errors.length) {setChangeTabAllowed(false)} else {setChangeTabAllowed(true)}
+    // },[errors])
```
luizakp, line 171: This is a TODO or just a leftover from the previous version? I would add a comment if it's a todo or delete this

PR #174, opened 2023-08-01
Refactor CustomEvents out

ribeirojose:


apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -30,51 +30,54 @@ const POOL_TYPES_MAPPER = {
   GyroE: "Gyro E-CLP",
 };
 
+export enum PoolSimulatorFormTabs {
```
ribeirojose, line 33: use an enum to ensure we're always using the expect tab values when manipulating them

PR #179, opened 2023-08-03
[Pool Simulator ]Calculate gyroe derivate values based on API request

yvesfracari:


apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -1,14 +1,39 @@
...
+    body: JSON.stringify({
+      alpha: String(data.poolParams?.alpha),
+      beta: String(data.poolParams?.beta),
+      l: String(data.poolParams?.lambda),
+      c: String(data.poolParams?.c),
+      s: String(data.poolParams?.s),
+    }),
+  })
+    .then((res) => res.json())
+    .then((res) => res as DerivedGyroEParamsFromSubgraph);
```
ribeirojose, line 32: why do the await then `.then`?  `#could` do all in one Promise.

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -26,45 +51,39 @@ export function convertAnalysisDataToAMM(data: AnalysisData) {
...
-          ),
-          tokens: data.tokens.map((token) => ({
-            address: String(token.symbol), // math use address as key, but we will use symbol because custom token will not have address
-            balance: String(token.balance),
-            decimals: token.decimal,
-            priceRate: String(token.rate),
-          })),
-          tokensList: data.tokens.map((token) => String(token.symbol)),
+      return fetchECLPDerivativeParams(data)
+        .then((derivedParams) => {
```
ribeirojose, line 59: since this is an async function you can just await the fetch -- think we should commit to one pattern unless there's a specific reason for not doing it

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -26,45 +51,39 @@ export function convertAnalysisDataToAMM(data: AnalysisData) {
...
-          tokenRates: data.tokens.map((token) => String(token.rate)),
-        }),
-      );
+              gyroEParams: {
+                alpha: String(data.poolParams?.alpha),
+                beta: String(data.poolParams?.beta),
+                lambda: String(data.poolParams?.lambda),
+                c: String(data.poolParams?.c),
+                s: String(data.poolParams?.s),
+              },
```
ribeirojose, line 80: You're doing the same thing above, a potential refactor here

apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -166,14 +167,26 @@ export function PoolSimulatorProvider({ children }: PropsWithChildren) {
...
+        setInitialData
       );
     }
     if (pathname === "/poolsimulator/analysis") {
       setIsGraphLoading(false);
     }
   }, [pathname]);
 
+  useEffect(() => {
+    if (!initialData.poolParams) return;
```
ribeirojose, line 179: this is redundant considering the next line

ribeirojose (dismissed):
looking 🔪 !

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -1,14 +1,39 @@
...
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      alpha: String(data.poolParams?.alpha),
+      beta: String(data.poolParams?.beta),
+      l: String(data.poolParams?.lambda),
+      c: String(data.poolParams?.c),
+      s: String(data.poolParams?.s),
```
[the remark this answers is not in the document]
yvesfracari: The two things here are the lambda, which the API expects just `l`, and transforming to strings. Regarding the `lambda` I thought that wouldn't be good to change the API pattern or our pattern (that is the same as the subgraph). I think that they changed the API bc `lambda` is reserved in Python.

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
yvesfracari: Is almost the same, here we don't rename the `lambda`

PR #180, opened 2023-08-03
[Pool simulator] Add dynamic parameters on forms menu

yvesfracari:
https://www.loom.com/share/de00909e2457404c91d7f7d0adeaf853?sid=54d9b169-2c25-4674-8961-26958d26fb33


apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -232,6 +231,102 @@ function SearchPoolFormWithDataForm({
   );
 }
 
+export function AnalysisMenu() {
```
ribeirojose, line 234: Why move this component to here? 😅

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -232,6 +231,102 @@ function SearchPoolFormWithDataForm({
...
+              }}
+              defaultValue={"0"}
+            >
+              {initialData?.tokens.map(({ symbol }, index) => (
+                <SelectItem key={symbol} value={index.toString()}>
+                  {symbol}
+                </SelectItem>
+              ))}
+            </Select>
+          </div>
```
ribeirojose, line 285: This is just a reiteration of the former component with the form underneath it right? Couldn't we just pass the select as a prop to `IndexMenu` and render it there when we need it to avoid repeating all this code?

apps/balancer-tools/src/app/poolsimulator/(components)/PoolParamsForm.tsx
```diff
@@ -141,13 +141,13 @@ const inputMapper = {
 
 const createPayload = (
   poolType: keyof typeof inputMapper,
-  fieldData: FieldValues,
+  fieldData: FieldValues
```
ribeirojose, line 144: prettier format

apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -128,6 +128,14 @@ export function PoolSimulatorProvider({ children }: PropsWithChildren) {
     if (token) setCurrentTabToken(token);
   }
 
+  async function asyncSetAMM(
```
ribeirojose, line 131: do we need loading states for these?

ribeirojose (dismissed):

apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
yvesfracari: Since the `IndexMenu` was here I wanted to let both here 😅

PR #181, opened 2023-08-03
[Pool Simulator] GyroE - Adapt swap curve to not go beyond swap limits

luizakp:
This PR:
- Renames StableCurve to SwapCurve
- Add swap limitations to be the pool balances
- Adapts de SwapCurve scale to the shorter curve
https://www.loom.com/share/3143cdb9ed964c0b8531b545e64a476f?sid=0f2ce497-43f5-4411-a7e6-c2dc2b7afefa


apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
```diff
@@ -9,7 +9,7 @@ import { Spinner } from "#/components/Spinner";
 import { TokensData, usePoolSimulator } from "#/contexts/PoolSimulatorContext";
 import { formatNumber } from "#/utils/formatNumber";
 
-import { calculateCurvePoints } from "./StableCurve";
+import { calculateCurvePoints } from "./SwapCurve";
```
yvesfracari, line 12: 👍

apps/balancer-tools/src/app/poolsimulator/(components)/SwapCurve.tsx
```diff
@@ -126,6 +142,47 @@ export function StableCurve() {
...
+      initial: Math.min(...initialAmountsOut),
+      custom: Math.min(...customAmountsOut),
+    };
+
+    const limits = {
+      lowerIn: Math.min(maxOfIn.initial, maxOfIn.custom),
+      higherOut: Math.max(minOfOut.initial, minOfOut.custom),
+    };
+
+    if (maxOfIn.initial === maxOfIn.custom) {
```
yvesfracari, line 170: Did you use this condition for the case of two metastable pools, right? If so, this will not work for all the cases.

apps/balancer-tools/src/app/poolsimulator/(components)/SwapCurve.tsx
```diff
@@ -126,6 +142,47 @@ export function StableCurve() {
     ),
   ];
 
+  function getGraphScale({
```
yvesfracari, line 145: `#should` use a simpler heuristic such as the token balances itself.

apps/balancer-tools/src/app/poolsimulator/(components)/SwapCurve.tsx
yvesfracari: e.g. two metastable pools but with different tokens balances.

ribeirojose (commented):
Looks good, otherwise defer to @yvesfracari comments

PR #183, opened 2023-08-04
[Pool Simulator] Adapt impact curve to not go beyond CLP limits

luizakp:
This PR:
- Add Impact price limitations to be within the pool balances. So Amount respects the pool balances and the price impact is showing accordingly 
- Adapts de ImpactCurve scale to the shorter curve on X = limit of X
- Adapts de ImpactCurve scale to the biggest Y on the limit of X


apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -168,3 +168,24 @@ export function convertGqlToAnalysisData(poolData: PoolQuery): AnalysisData {
...
 }
+
+export function calculateCurvePoints({
+  balance,
+  start = 0,
+}: {
+  balance?: number;
+  start?: number;
+}) {
+  if (!balance || typeof start == "undefined") return [];
```
devjoaov, line 179: could be `!balance || start === undefined`

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
```diff
@@ -156,6 +171,24 @@ export function ImpactCurve() {
     ),
   ];
 
+  const maxOfInitial = Math.max(
```
ribeirojose, line 174:
```suggestion
  const maxFromInitialAmounts = Math.max(
```

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
```diff
@@ -156,6 +171,24 @@ export function ImpactCurve() {
...
+    ...initialAmountsAnalysisTokenIn,
+    ...initialAmountsTabTokenIn,
+  );
+  const maxOfCustom = Math.max(
+    ...variantAmountsAnalysisTokenIn,
+    ...variantAmountsTabTokenIn,
+  );
+  const xlimit = Math.min(maxOfInitial, maxOfCustom);
+  const indexOfXLimit = initialAmountsAnalysisTokenIn.indexOf(xlimit);
```
ribeirojose, line 183:
```suggestion
  const xLimitIndex = initialAmountsAnalysisTokenIn.indexOf(xlimit);
```
also, would this always yield a result or could it be `-1`? 🤔 xLimit could be a value that's not in `initialAmountsAnalysisTokenIn` right?

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
```diff
@@ -168,3 +168,24 @@ export function convertGqlToAnalysisData(poolData: PoolQuery): AnalysisData {
       };
   }
 }
+
+export function calculateCurvePoints({
```
ribeirojose, line 172: could add a test here?

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
luizakp: you are right! fixing rn

apps/balancer-tools/src/app/poolsimulator/(utils)/index.ts
luizakp: Done :) https://github.com/bleu-studio/balancer-tools/pull/183/commits/075a767bd7d68e7e5e837fc7e7da802ffae8ae14

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
luizakp: done here https://github.com/bleu-studio/balancer-tools/pull/183/commits/1beba2c801689c8e75d149f14dfd3ad4834829dc

ribeirojose (approved):
👌

apps/balancer-tools/src/__tests__/utils/calculateCurvePoints.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const result = calculateCurvePoints({ balance, start });
+
+    expect(result.length).toBe(121);
+
+    expect(result[0]).toBe(start);
+  });
+
+  it("should return an empty array when balance is not provided", () => {
+    const result = calculateCurvePoints({});
+    expect(result).toEqual([]);
```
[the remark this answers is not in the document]
luizakp: we return an empty array because with this the `<Plot>` still renders but with empty data instead of breaking it

apps/balancer-tools/src/__tests__/utils/calculateCurvePoints.test.tsx
```diff
@@ -0,0 +1,44 @@
+import { calculateCurvePoints } from "#/app/poolsimulator/(utils)";
+
+describe("calculateCurvePoints", () => {
+  it("should return an array of calculated points with a given balance and start", () => {
+    const balance = 1000;
+    const start = 10;
+    const result = calculateCurvePoints({ balance, start });
+
+    expect(result.length).toBe(121);
```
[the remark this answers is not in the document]
luizakp: great question actually, we defined 
```
const numberOfPoints = 100
```
but then used 

```
{ length: numberOfPoints + 20 },
```

So 121 is 120 `numberOfPoints` + the start point.

I would say it's just the number of points we decided to have, but I don't know if there is a deeper reason @yvesfracari

PR #184, opened 2023-08-07
Fix eslint errors

ribeirojose:


yvesfracari (approved):
Checked at Pool Simulator-related files and LGTM! Just a question, is the `type` on the imports a style preference or is it related to something else?

ribeirojose:
@yvesfracari It's related to the code that's going to be emitted after transpiling: https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/

luizakp (changes requested):
I'm checking what is happening exactly, but internal manager is not showing toasts or the loading buttons

PR #192, opened 2023-08-09
show bal price for each round

ribeirojose:


apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default async function BALPrice({
+  roundId,
+}: {
+  roundId: string | string[];
+}) {
+  invariant(!Array.isArray(roundId), "roundId cannot be a list");
+  const round = Round.getRoundByNumber(roundId);
+
+  return <Price data={getBALPriceByRound(round)} />;
```
ribeirojose, line 14: doing this so we can use `Suspense` and show it as loading

apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { getBALPriceByRound } from "../(utils)/getBALPriceByRound";
+import { Round } from "../(utils)/rounds";
+
+export default async function BALPrice({
+  roundId,
+}: {
+  roundId: string | string[];
+}) {
+  invariant(!Array.isArray(roundId), "roundId cannot be a list");
```
luizakp, line 11: this is what we had on the Remix project to early return? should we refactor the other apps to start using this?

apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/round/[roundId]/page.tsx
```diff
@@ -1,13 +1,20 @@
...
+}) {
+  const { network, poolId, roundId } = params;
 
   return (
-    <div className="flex h-full w-full flex-col justify-center rounded-3xl">
+    <div className="flex h-full w-full flex-col justify-center rounded-3xl text-white">
       hello from pool {poolId} in network {network} for round {roundId}
+      <Suspense fallback={"Loading..."}>
+        <BALPrice roundId={roundId} />
+      </Suspense>
```
Comment (luizakp) [inline apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/round/[roundId]/page.tsx:17]: 👏

apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
ribeirojose: yup. @luizakp do you think it's worth the time going back and refactoring the other apps to do this? what do we gain by doing it?

PR #193, opened 2023-08-09
Rawallon/bal 610 list pools in round page

Rawallon:
This Branch does three things:
 - Initial DuneAPI impl.
 - Get pools info for a given round
 - Simple styling for displaying card info


apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
```diff
@@ -1,13 +1,58 @@
...
+import PoolCard from "../(components)/PoolCard";
 
 export default function Page() {
+  const [poolsData, setPoolsData] = useState<Row[]>([]);
+  const [hasError, setHasError] = useState("");
   const { roundId } = useParams();
+  useEffect(() => {
+    if (!poolsData.length) {
+      const fetchData = async () => {
+        const poolsDataRows = await (
```
Comment (Rawallon) [inline apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:18]: This made me realise maybe this component shouldnt be client

apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts
```diff
@@ -0,0 +1,15 @@
+import { NextRequest, NextResponse } from "next/server";
+
+import { DuneAPI } from "#/lib/dune";
+import { voteGaugeByID } from "../../mock_apis";
+
+export async function GET(
```
Comment (Rawallon) [inline apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts:6]: Not sure about this, did this way since we dont want to use DuneAPI on the client

apps/balancer-tools/src/lib/dune.ts
```diff
@@ -0,0 +1,34 @@
...
+export class DuneAPI {
+  private client: DuneClient;
+
+  constructor() {
+    this.client = new DuneClient(duneApiKey ?? "");
+  }
+
+  async getPoolsByRoundId(roundId: number = 0): Promise<Row[] | ErrorReturn> {
+    invariant(roundId > 0, "An valid roundId must be passed");
+    const parameters = [QueryParameter.text("round_id", String(roundId))];
```
Rawallon, line 27: Maybe also add limit on this query so we dont show the 100+ pools

apps/balancer-tools/src/app/apr/round/(components)/PoolCard.tsx
```diff
@@ -0,0 +1,29 @@
+import { ChevronRightIcon } from "@radix-ui/react-icons";
```
Rawallon, line 1: Not sure if I'm using the folder structure correctly

apps/balancer-tools/src/app/apr/round/(components)/PoolCard.tsx
Rawallon, line 1: Not sure if I'm using the folder structure correctly

apps/balancer-tools/package.json
```diff
@@ -23,6 +23,7 @@
     "@bleu-balancer-tools/math-poolsimulator": "workspace:*",
     "@bleu-balancer-tools/math-stableswapsimulator": "workspace:*",
     "@bleu-balancer-tools/tsconfig": "workspace:*",
+    "@cowprotocol/ts-dune-client": "^0.0.2",
```
Rawallon, line 26: Not strictly necessary, didn't want to have to re-implement pooling logic

Rawallon:
Still having trouble with linting, [This suggestion](https://stackoverflow.com/a/67273107) seems to have fixed it, could anyone test if it would keep working if we made this change?

apps/balancer-tools/src/app/apr/mock_apis.tsx
```diff
@@ -0,0 +1,147 @@
+export const voteGaugeByID = [
```
ribeirojose, line 1: Could be a simple `ts` or JSON

apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts
Comment (ribeirojose) [reply apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts:6]: Got it, that makes sense! 

I think that's the default way to reason about this -- but we could could also leverage server components like I did in my PR to load it (I especially enjoy it because I can forget about local loading states, just leaving `Suspense` handle that 😅 )

apps/balancer-tools/src/lib/dune.ts
```diff
@@ -0,0 +1,34 @@
+import "server-only";
```
ribeirojose, line 1: 👌

apps/balancer-tools/src/lib/dune.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor() {
+    this.client = new DuneClient(duneApiKey ?? "");
+  }
+
+  async getPoolsByRoundId(roundId: number = 0): Promise<Row[] | ErrorReturn> {
+    invariant(roundId > 0, "An valid roundId must be passed");
+    const parameters = [QueryParameter.text("round_id", String(roundId))];
+
+    return this.client
+      .refresh(2834602, parameters)
```
ribeirojose, line 30: I suppose this is the query ID? Must be marked as such

apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
Comment (ribeirojose) [reply apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:18]: I think there's nothing wrong in doing it this way -- but think we'll end up needing to write less code if we do the server components approach. Also we'll learn something new 😉

apps/balancer-tools/src/app/apr/round/(components)/PoolCard.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { formatNumber } from "#/utils/formatNumber";
+
+export default function PoolCard({ symbol, pctVotes, numVotes }) {
```
ribeirojose, line 5: `#must` have TS types

ribeirojose:
> Still having trouble with linting, [This suggestion](https://stackoverflow.com/a/67273107) seems to have fixed it, could anyone test if it would keep working if we made this change?

feel free to make it and merge this -- it should work cross platform

ribeirojose (approved):

PR #196, opened 2023-08-09
[Pool Simulator] Add Depth cost for Gyro 2CLP and calculateDepthCost test

luizakp:
image

This PR:
- refactor `calculateDepthCost()` to reduce duplication of logic
- add case `PoolTypeEnum.Gyro2` to `calculateDepthCost()`
- add test to `calculateDepthCost()`


apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
```diff
@@ -231,47 +231,57 @@ const createDataObject = (
...
   data: AnalysisData,
   amm: AMM<PoolPairData>,
   poolType: PoolTypeEnum,
 ) {
-  const { analysisToken } = usePoolSimulator();
-
+  const analysisToken = data.tokens.find(
+    (token) => token.symbol !== pairToken.symbol,
+  );
+  if (!analysisToken) throw new Error("Analysis token not found");
```
yvesfracari, line 244: If you did this to not have to use the context, I think that would be best to require the analysis token as an input of the function.

apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
```diff
@@ -231,47 +231,57 @@ const createDataObject = (
...
+  const alphaBeta = {
+    alpha:
+      poolType === PoolTypeEnum.GyroE
+        ? data.poolParams?.alpha
+        : (data.poolParams?.sqrtAlpha as number) ** 2,
+    beta:
+      poolType === PoolTypeEnum.GyroE
+        ? data.poolParams?.beta
+        : (data.poolParams?.sqrtBeta as number) ** 2,
+  };
```
yvesfracari, line 270: would use two variables instead of a dict here

yvesfracari (approved):

apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
luizakp: you say this in case of more than 2 tokens right?

PR #198, opened 2023-08-09
[APR] Populate data on pool page

Rawallon:
I would appreciate opinions on the approach I've taken. The problem I aimed to solve was that all the information about which network/round the user has selected was previously managed by the Layout.tsx. I wanted components inside to be able to access some of that information. I chose to use context since it is the most straightforward way to retain and share data.

### DIFF
--- a/apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
+++ b/apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
@@ -36,8 +36,10 @@ function PoolCard({
 
 export default async function PoolsCards({
   data,
+  onClick,
 }: {
   data: ReturnType<typeof fetcher<DunePoolData[] | { error: string }>>;
+  onClick: (poolId: string) => void;
 }) {
   const poolsData = await data;
 
@@ -46,7 +48,9 @@ export default async function PoolsCards({
   return (
     <div className="space-y-6 w-full">
       {poolsData.map((poolData) => (
-        <PoolCard data={poolData} />
+        <div onClick={() => onClick(poolData.symbol)}>
+          <PoolCard data={poolData} />
+        </div>
       ))}
     </div>
   );
--- a/apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
+++ b/apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
@@ -3,6 +3,7 @@
 import { Suspense } from "react";
 
 import Loading from "#/app/metadata/[network]/pool/[poolId]/loading";
+import { useSelectedPoolRoundContext } from "#/contexts/SelectedPoolRound";
 import { fetcher } from "#/utils/fetcher";
 
 import PoolsCards from "../(components)/PoolsCards";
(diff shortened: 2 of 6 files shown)


apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
```diff
@@ -12,6 +13,12 @@ export default function Page({
 }: {
   params: { roundId: string };
 }) {
+  const { setSelectedPool } = useSelectedPoolRoundContext();
+
+  function onClick(poolId: string) {
+    setSelectedPool(poolId);
+  }
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:20]: this function definition doesn't add much, does it?

apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
```diff
@@ -46,7 +48,9 @@ export default async function PoolsCards({
   return (
     <div className="space-y-6 w-full">
       {poolsData.map((poolData) => (
-        <PoolCard data={poolData} />
+        <div onClick={() => onClick(poolData.symbol)}>
+          <PoolCard data={poolData} />
+        </div>
```
ribeirojose, line 53: couldn't this simply be a link to the pool page?

apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
```diff
@@ -12,6 +13,12 @@ export default function Page({
 }: {
   params: { roundId: string };
 }) {
+  const { setSelectedPool } = useSelectedPoolRoundContext();
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:16]: wouldn't we already be setting the selected pool in the path? How an additional context would be helpful here?

apps/balancer-tools/src/contexts/SelectedPoolRound.tsx
```diff
@@ -0,0 +1,97 @@
...
+  children: ReactNode;
+}
+
+export const SelectedPoolRoundContextProvider = ({
+  children,
+}: RoundContextProviderProps) => {
+  const router = useRouter();
+  const [selectedRound, setSelectedRound] = useState<string | undefined>();
+  const [selectedPool, setSelectedPool] = useState<string | undefined>();
+  const [selectedNetwork, setNetworkId] = useState<string>("1");
```
ribeirojose, line 38: we already have this defined in the params, what do we gain by adding this complication?

ribeirojose (changes requested):
not sure what this PR is adding apart from the navigation from the card to the pool page -- did you add any new pool data to the pool page that maybe you forgot to push? 

let me know if you think I might be missing something but I don't think this is ready to merge yet

apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
Comment (luizakp) [reply apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:16]: you mean instead of using `setSelectedPool` we could just redirect to the pool url?

Comment (ribeirojose) [reply apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:16]: yup @luizakp

PR #199, opened 2023-08-09
[Pool simulator] Fix depth cost not converging

yvesfracari:
This PR:

- Add checking on the `_tokenInForExactSpotPriceAfterSwap` don`t guess values outside the pool range
- Add backtracking line search variation on the Newton method of the same function to avoid steps that don't get closer to the desired result
- Add spot price alpha and beta transformation for checking if the new spot price is reachable.
- Adapt spot price precision depending on the spot price value.


apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
```diff
@@ -263,14 +265,19 @@ function calculateDepthCost(
...
+              spotPricePrecision,
             ),
             type: "2% of price change",
           };
     case PoolTypeEnum.GyroE: {
       // For CLP pools we'll assume depth cost as the pool depth == 99% of the liquidity or 2% of the current spot price (if possible)
+      const newSpotPriceOnAlphaAndBetaBase =
+        tokenIn.symbol === data.tokens[1].symbol
+          ? newSpotPrice
+          : 1 / newSpotPrice;
```
ribeirojose, line 277: 👀 why?

packages/math-poolsimulator/src/index.ts
```diff
@@ -75,9 +75,43 @@ export class AMM<TPoolPairData extends PoolPairData> {
...
+
+    // Some pools are limited, so this iteration is to avoid a guess out of pool range.
+    for (let i = 0; i < 20; i++) {
+      if (this.math._checkIfInIsOnLimit(poolPairData, newInGuess)) break;
+      inGuessDelta = inGuessDelta.div(bnum(2));
+      newInGuess = inGuessValue.plus(inGuessDelta);
+    }
+
+    // This is the backtracking line search variation of the Newton-Raphson method
+    // that avoids guesses that not significantly improve the result
```
ribeirojose, line 90:
```suggestion
    // that avoids guesses that don't significantly improve the result
```

apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
yvesfracari: Just add a comment about it:
```
      // The Alpha and Beta values are considering token 0 in units of token 1.
      // This means that token 1 must be the the tokenIn
      // And token 0 must be the tokenOut
```

packages/math-poolsimulator/src/index.ts
```diff
@@ -75,9 +75,43 @@ export class AMM<TPoolPairData extends PoolPairData> {
...
+    for (let i = 0; i < 20; i++) {
+      if (this.math._checkIfInIsOnLimit(poolPairData, newInGuess)) break;
+      inGuessDelta = inGuessDelta.div(bnum(2));
+      newInGuess = inGuessValue.plus(inGuessDelta);
+    }
+
+    // This is the backtracking line search variation of the Newton-Raphson method
+    // that avoids guesses that don't significantly improve the result
+    const alpha = bnum(2);
+    const beta = bnum(8);
```
ribeirojose, line 92: is there anything special about these alpha/beta values? I assume they're not the same values from the pool params we're simulating, but rather only here for the numerical analysis part, right?

ribeirojose (dismissed):
Looking good! -- but I need to learn more numerical calculus 😶‍🌫️

packages/math-poolsimulator/src/index.ts
yvesfracari: I will add a comment about this, but the alpha and beta are from the algorithm. You have to set the alpha can be [2,infinity] and beta [1, infinity]

apps/balancer-tools/src/app/poolsimulator/(components)/DepthCost.tsx
```diff
@@ -246,18 +246,22 @@ export function calculateDepthCost(
   if (!analysisToken) throw new Error("Analysis token not found");
   const tokenIn = poolSide === "in" ? analysisToken : pairToken;
   const tokenOut = poolSide === "in" ? pairToken : analysisToken;
-  const newSpotPrice = amm.spotPrice(tokenIn.symbol, tokenOut.symbol) * 1.02;
+  const currentSpotPrice = amm.spotPrice(tokenIn.symbol, tokenOut.symbol);
+  const newSpotPrice = currentSpotPrice * 1.02;
+  const spotPricePrecision = currentSpotPrice * 0.0002;
```
luizakp, line 251: just noticed that this is not being used. Is that correct?

yvesfracari: The spotPricePrecision? I forgot to add when merging the branches but now it is being used, right?

luizakp: yesss now I synced with main hahaha

PR #200, opened 2023-08-10
[APR] add pool pct votes to pool/{id}/round/{number}

ribeirojose:
- refactor filtering logic from PoolSearch component;
- copy voting-gauges from https://github.com/balancer/frontend-v2 (@mendesfabio do you know where they store the script to update these voting gauges?)
- add pool pct votes to pool/${id}/round/${number} page, to be used to compute the pool APR in a certain round
- downgrade Next to 13.3.4 to attempt to make it work on Netlify again 🤡  issue: https://answers.netlify.com/t/server-edge-not-defined-error-on-nextjs-ssr-functions-cause-site-to-return-500-errors/91793

### DIFF
--- a/apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
+++ b/apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
@@ -6,10 +6,11 @@ import { Round } from "../(utils)/rounds";
 export default async function BALPrice({
   roundId,
 }: {
-  roundId: string | string[];
+  roundId?: string | string[];
 }) {
   invariant(!Array.isArray(roundId), "roundId cannot be a list");
-  const round = Round.getRoundByNumber(roundId);
+
+  const round = roundId? Round.getRoundByNumber(roundId) : Round.currentRound();
 
   return <Price data={getBALPriceByRound(round)} />;
 }
@@ -24,10 +25,10 @@ export const Price = async ({
   return (
     <div>
       BAL price:{" "}
-      {price.toLocaleString("en-US", {
+      {price ? price.toLocaleString("en-US", {
         style: "currency",
         currency: "USD",
-      })}
+      }) : "error"}
     </div>
   );
 };
--- a/apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
+++ b/apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
@@ -10,7 +10,9 @@ export default async function BALPrice({
 }) {
   invariant(!Array.isArray(roundId), "roundId cannot be a list");
 
-  const round = roundId? Round.getRoundByNumber(roundId) : Round.currentRound();
+  const round = roundId
+    ? Round.getRoundByNumber(roundId)
+    : Round.currentRound();
 
   return <Price data={getBALPriceByRound(round)} />;
 }
@@ -25,10 +27,12 @@ export const Price = async ({
   return (
     <div>
       BAL price:{" "}
-      {price ? price.toLocaleString("en-US", {
-        style: "currency",
-        currency: "USD",
-      }) : "error"}
+      {price
+        ? price.toLocaleString("en-US", {
+            style: "currency",
+            currency: "USD",
+          })
+        : "error"}
     </div>
   );
 };
--- a/apps/balancer-tools/src/app/apr/(components)/PoolPctVotesInRound.tsx
+++ b/apps/balancer-tools/src/app/apr/(components)/PoolPctVotesInRound.tsx
@@ -14,7 +14,10 @@ export default async function PoolPctVotesInRound({
   invariant(!Array.isArray(roundId), "roundId cannot be a list");
   invariant(!Array.isArray(poolId), "poolId cannot be a list");
 
-  const gaugeData = await fetcher<DuneGaugeData[] | { error: string }>(`http://localhost:3000/apr/rounds/${roundId}`, {method: "GET"});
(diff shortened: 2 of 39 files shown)


apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
```diff
@@ -6,10 +6,13 @@ import { Round } from "../(utils)/rounds";
 export default async function BALPrice({
   roundId,
 }: {
-  roundId: string | string[];
+  roundId?: string | string[];
```
Rawallon, line 9: Did this arg became optional because of the changes on `apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/page.tsx`?

apps/balancer-tools/src/app/apr/(utils)/rounds.ts
```diff
@@ -6,10 +6,10 @@ export class Round {
   value: string;
   activeRound: boolean;
 
-  constructor(endDate: Date, roundNumber: number, active: boolean) {
+  constructor(endDate: Date, roundNumber: number) {
     this.endDate = endDate;
     this.value = String(roundNumber);
-    this.activeRound = active;
+    this.activeRound = endDate.getTime() > Date.now();
```
Rawallon, line 12: Clever

apps/balancer-tools/src/app/apr/(components)/BALPrice.tsx
ribeirojose: Yes -- if we don't use a round it will just assume the current time

apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/page.tsx
Comment (Rawallon) [inline apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/page.tsx:1]: I'm assuming we always want that the client has a `/<roundnumber>`, similar to what we do in the root app I think he should be redirected to `/pool/<id>/round/current`

apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
```diff
@@ -1,52 +1,62 @@
...
+            <div className="text-white font-bold text-lg mb-2">{poolId}</div>
+          </div>
+          <div className="flex items-center">
+            <div className="text-sm">
+              <p className="text-white leading-none mb-1">
+                {(pctVotes * 100).toFixed(2)}% Voted
+              </p>
+              <p className="text-white text-xs">{formatNumber(votes)} Votes</p>
+              <Suspense fallback={"Loading..."}>
+                <BALPrice />
```
Rawallon, line 35: Isn't this going to display current value regardless which round is selected?

apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/page.tsx
Comment (ribeirojose) [reply apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/page.tsx:1]: not really -- we want them to use the select to do this

apps/balancer-tools/src/app/apr/(components)/PoolPctVotesInRound.tsx
```diff
@@ -0,0 +1,47 @@
...
+  poolId,
+}: {
+  roundId?: string | string[];
+  poolId: string | string[];
+}) {
+  invariant(!Array.isArray(roundId), "roundId cannot be a list");
+  invariant(!Array.isArray(poolId), "poolId cannot be a list");
+
+  const gaugeData = await fetcher<DuneGaugeData[] | { error: string }>(
+    `http://localhost:3000/apr/rounds/${roundId}`,
```
devjoaov, line 18: shoudn't this be in the .env?

apps/balancer-tools/src/components/SearchPoolForm/index.tsx
Rawallon, line 1: One thing I've showed to Lu but I'm not 100% you're aware is that for some reason the onBlur on the Input in this component isn't working. This needs to be fixed but currently is low prio

apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
ribeirojose: ops, good catch!

apps/balancer-tools/src/app/apr/(components)/PoolPctVotesInRound.tsx
ribeirojose: good catch! Not in the .env but it should be using the URL host from the request instead of doing this

apps/balancer-tools/src/components/SearchPoolForm/filterPoolInput.ts
```diff
@@ -0,0 +1,64 @@
...
+      for (const subValue of Object.values(item)) {
+        if (typeof subValue === "string" && regex.test(subValue)) {
+          return true;
+        }
+      }
+    }
+  }
+
+  return false;
+}
```
ribeirojose, line 30: could be cleaner

apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx
```diff
@@ -20,7 +17,7 @@ export default function Page({
         </div>
       }
     >
-      <PoolsCards data={fetcher(`/apr/rounds/${roundId}`)} />
+      <PoolsCards roundId={roundId} />
```
Comment (devjoaov) [inline apps/balancer-tools/src/app/apr/round/[roundId]/page.tsx:20]: 👍

devjoaov (approved):
LGTM

PR #203, opened 2023-08-11
Rawallon/bal 615 calculate apr for a given round

Rawallon:
I considered calculating the APR using the endpoint `/apr/rounds/<RoundNumber>/apr?votingShare=123`, but then I realized that it would be highly inefficient. Therefore, I relocated it within the same process of fetching the pool data using `/apr/rounds/<RoundNumber>`. However, I'm not completely convinced about this approach as I believe it might significantly slow down the initial rendering. Therefore, I'm open to suggestions.

P.S.: The pages under `.../apr/rounds/[roundId]/...` will not be merged if everyone agrees to maintain the APR calculation on `/apr/rounds/<RoundNumber>`.


apps/balancer-tools/src/app/apr/(utils)/getBalEmission.ts
```diff
@@ -0,0 +1,28 @@
+interface BalEmissions {
+  year: number;
+  weekly: number;
+  total_supply: number;
+}
+
+export const getBalEmissions = (year: number): BalEmissions => {
```
ribeirojose, line 7: this could come from [BalancerSDK](https://github.com/balancer/balancer-sdk/blob/379883359a0dead100282158f7aa2d13c255a7a3/balancer-js/src/modules/data/bal/emissions.ts)

You can see how the defyltica folks implemented it [here](https://github.com/Xeonus/balancer-analytics/blob/8722b3776340a764494b11c16cdb4c6ef3bcdda0/src/pages/Emissions/index.tsx#L68)

ribeirojose: nvm -- that'd make us need to install ethers again 🤦 let's just keep as is

Rawallon: We could yoink just [emissions.ts](https://github.com/balancer/balancer-sdk/blob/379883359a0dead100282158f7aa2d13c255a7a3/balancer-js/src/modules/data/bal/emissions.ts)

ribeirojose: yea that's what I did here https://github.com/bleu-studio/balancer-tools/pull/205/commits/7947adf430cf49aaab1b060ec985dc20efd86217 😅

ribeirojose: cool implementation nevertheless -- but missing docs for how you came up with all that. See their excellent docs in their own implementation for reference

apps/balancer-tools/src/app/apr/rounds/[roundId]/apr/route.ts
```diff
@@ -0,0 +1,43 @@
...
+      52 *
+      balPrice) /
+      totalTvl) *
+    100;
+  return NextResponse.json({
+    apr: apr,
+    balPrice: balPrice,
+    balEmissions: balEmissions,
+    totalTvl: totalTvl,
+    ASD: currentRound.endDate.toISOString().slice(0, 10),
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/rounds/[roundId]/apr/route.ts:41]: what does this mean?

apps/balancer-tools/src/app/apr/rounds/[roundId]/apr/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  // const dune_request = await duneAPI.getTVLByRoundId(context.params.roundId)
+  const currentRound = Round.getRoundByNumber(context.params.roundId);
+  const dune_request = mockGetTVLByRoundId;
+  const totalTvl = dune_request[0]["total_tvl"];
+  const balPrice = await getBALPriceByRound(currentRound);
+
+  const balEmissions = getBalEmissions(currentRound.endDate.getFullYear())[
+    "weekly"
+  ];
+  // APR = emissions in week * voting share * weeks in year * BAL price / TVL
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/rounds/[roundId]/apr/route.ts:28]: missing reference quote.

apps/balancer-tools/src/lib/dune.ts
```diff
@@ -44,4 +51,28 @@ export class DuneAPI {
...
+    const queryId = 2842639;
+    const parameters = [
+      QueryParameter.number(
+        "End date",
+        roundData.endDate.toISOString().slice(0, 10),
+      ),
+    ];
+
+    return this.client
+      .refresh(queryId, parameters)
```
ribeirojose, line 71: #question do we really need to call refresh? Couldn't we use a cached version of that data from a result that's already run? I wonder if that would make us spend less credits.

apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
```diff
@@ -1,13 +1,19 @@
...
 
-import { DunePoolData } from "#/lib/dune";
 import { fetcher } from "#/utils/fetcher";
 import { formatNumber } from "#/utils/formatNumber";
 
+export interface PoolData {
+  pct_votes: number;
+  symbol: string;
+  votes: number;
+  apr: number;
```
ribeirojose, line 10: why did we need to define this here again?

ribeirojose (commented):
I get this gist of this, but where is the APR value being fetched/shown right now? Also, do we really need those endpoints? Couldn't we just use the same data fetching pattern we established in e.g. PoolsCards?

apps/balancer-tools/src/app/apr/mock_apis.tsx
```diff
@@ -1,3 +1,9 @@
+export const mockGetTVLByRoundId = [
+  {
+    total_tvl: 2128.221644270715,
+  },
```
ribeirojose, line 4: for whom? How are going to know whose TVL this is from?

apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts
```diff
@@ -1,15 +1,39 @@
...
+  )["weekly"];
+
+  requestJson.forEach((poolData) => {
+    // APR = emissions in week * voting share * weeks in year * BAL price / TVL
+    poolData["apr"] =
+      (balWeeklyEmissions * poolData.pct_votes * 52 * balPrice) /
+      totalTvl /
+      100;
+    return poolData;
+  });
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts:37]: do we even need an APR endpoint and that same logic implemented here?

Comment (ribeirojose) [reply apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts:37]: why are you plugging a random TVL to every pool in that list?

ribeirojose (changes requested):
I don't think this is ready to merge -- this is at most a draft PR but there are several questions left yet. Happy to go through the issues together

apps/balancer-tools/src/lib/dune.ts
Rawallon: This should be cacheable, I have to look into how Next does it.

I think the a `import { cache } from 'react'` should sufice, but not 100% sure on that

apps/balancer-tools/src/app/apr/round/(components)/PoolsCards.tsx
Rawallon: I think placing this here will make it easier when displaying data on the rounds page

apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts
Comment (Rawallon) [reply apps/balancer-tools/src/app/apr/rounds/[roundId]/route.ts:37]: > do we even need an APR endpoint and that same logic implemented here?

Not necessarily. I began developing it as its own route (which is why the files are present). However, considering that we would want to retrieve the APR when fetching the pools, I thought it might be more convenient to return everything within a single endpoint.

> why are you plugging a random TVL to every pool in that list?

That's another aspect I'm not entirely certain about how to approach. Querying for individual pools is inefficient, and querying for all pools is expensive. The only solution I can currently think of is to perform a query for a list (using SQL IN), but I haven't tested that yet.

ribeirojose:
left a few ideas here https://github.com/bleu-studio/balancer-tools/pull/205

PR #204, opened 2023-08-11
Fix Gyro E 95 percent test

yvesfracari:


ribeirojose (commented):
😬 can't opine with a bit more context

luizakp (approved):
We can talk later so I can understand what you did hahahh but the tests are working

PR #207, opened 2023-08-14
[Pool Simulator] Implement FX Math Pt.1

yvesfracari:
This PR:
- Add the first version of spot price function for FX pools;
- Enable FX analysis without depth cost chart and with issues on tokens with different prices (e.g. EURS - USDC)
- Fix issue with graph scale on Swap Curve for FX pool.
- Enable 0 of swap fee.

### DIFF
--- a/apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
+++ b/apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
@@ -1,6 +1,5 @@
 import { MagnifyingGlassIcon } from "@radix-ui/react-icons";
 import * as Separator from "@radix-ui/react-separator";
-// import { usePathname, useRouter } from "next/navigation";
 import { ReactElement, useEffect, useState } from "react";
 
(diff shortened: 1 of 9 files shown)


apps/balancer-tools/src/app/poolsimulator/(components)/Menu.tsx
```diff
@@ -92,13 +90,13 @@ function IndexMenu() {
               }
               defaultValue={initialData}
               onTabChanged={(data) => {
-                if (!customData.poolParams?.swapFee) {
+                if (typeof customData.poolParams?.swapFee === "undefined") {
```
luizakp, line 93: typescript is weird kkkkkk why call a type as a string

luizakp (approved):
I would say we merge this PR and create another issue to address the spot price for tokens with different prices. 

We should have in mind that our main goal with this task wasn't achieved but we would rather divide it into smaller tasks than try to make it all here. What we achieved with this PR was implementing the Fx math structure and rendering something on the charts even if it's not always correct. WDYT?

PR #208, opened 2023-08-14
[Pool Simulator] Add web workers for swap and impact curve

luizakp:
This PR:
- Add `swap-curve-calculation.ts` and it's use on `SwapCurve`
- Add `impact-curve-calculation.ts` and it's use on `ImpactCurve`
- remove Spinners from the analysis page

https://www.loom.com/share/490f64ccec6d4b4aa11da8c096901806?sid=5b1de0d5-9e1f-4996-816b-3185307921f9

### DIFF
--- a/apps/balancer-tools/src/app/poolsimulator/(components)/AnalysisPage.tsx
+++ b/apps/balancer-tools/src/app/poolsimulator/(components)/AnalysisPage.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import { Spinner } from "#/components/Spinner";
 import { Tabs } from "#/components/Tabs";
 import { usePoolSimulator } from "#/contexts/PoolSimulatorContext";
 
@@ -14,10 +13,6 @@ import { TokensDistribution } from "./TokensDistribution";
 export default function Page() {
   const { initialData, analysisToken, setCurrentTabTokenByIndex } =
     usePoolSimulator();
-
-  if (!analysisToken.symbol) {
-    return <Spinner />;
-  }
 
   const indexCurrentTabToken = initialData?.tokens.findIndex(
     ({ symbol }) => symbol.toLowerCase() !== analysisToken.symbol.toLowerCase(),
--- a/apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
+++ b/apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
@@ -28,23 +28,21 @@ interface Amounts {
 const createAndPostWorker = (
   messageData: ImpactWorkerInputData,
   setInitialAmounts: Dispatch<SetStateAction<Amounts>>,
-  setCustomAmounts: Dispatch<SetStateAction<Amounts>>
+  setCustomAmounts: Dispatch<SetStateAction<Amounts>>,
 ) => {
   const worker = new Worker(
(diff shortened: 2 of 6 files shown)


luizakp:
Chore to do the web worker to depth cost and swap simulation is logged [here](https://linear.app/bleu-llc/issue/BAL-630/add-web-worker-to-swap-simulation-and-depth-cost)

These two views don't need computational power, but passing the amm as props does not provide access to it as a class, which disables us to use all the math on the workers 😓 

Therefore, I took the decision to create the AMM inside the worker, which make the states `initialAMM` and `customAMM` not necessary anymore if we also implement a web worker to depth cost and swap simulation

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
```diff
@@ -1,72 +1,101 @@
...
+  const { analysisToken, currentTabToken, initialData, customData } =
+    usePoolSimulator();
+
+  const [initialAmounts, setInitialAmounts] = useState<Amounts>({});
+  const [customAmounts, setCustomAmounts] = useState<Amounts>({});
+
+  if (!initialData || !customData) return <Spinner />;
 
-  if (!initialAMM || !customAMM) return <Spinner />;
+  useEffect(() => {
```
ribeirojose, line 37: `#should` extract this worker calling logic to outside of the component

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
```diff
@@ -1,72 +1,101 @@
...
+    };
+
+    worker.postMessage({
+      tokenIn: analysisToken,
+      tokenOut: currentTabToken,
+      data: initialData,
+      poolType: initialData.poolType,
+      swapDirection: "in",
+      type: "initial",
+    } as ImpactWorkerInputData);
```
ribeirojose, line 67: this way, you're posting multiple messages to the same worker which will end up making it queue them. You could instead instantiate multiple webworkers to achieve parallel computation given that results don't depend on one another.

ribeirojose: example here https://github.com/bleu-studio/balancer-tools/pull/209

ribeirojose (dismissed):
👌  great initiative!

apps/balancer-tools/src/app/poolsimulator/(components)/ImpactCurve.tsx
luizakp: got it! thanks! will do the same for swap curve

apps/balancer-tools/src/app/poolsimulator/(components)/AnalysisPage.tsx
```diff
@@ -1,6 +1,5 @@
 "use client";
```
yvesfracari, line 1: Should we move this to the `index` page file now that all the app is on a single page?

yvesfracari (dismissed):

apps/balancer-tools/src/app/poolsimulator/(components)/AnalysisPage.tsx
luizakp: I would say this is still a component. It would be too big to be on `index` IMO

PR #209, opened 2023-08-14
run parallel workers

ribeirojose:


luizakp (approved):
Thanks so much 😊 @ribeirojose

PR #215, opened 2023-08-16
[Pool Simulator] Fix share simulation feature

yvesfracari:
Closes BAL-575, BAL-620

### DIFF
--- a/apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
+++ b/apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
@@ -16,7 +16,7 @@ export function ShareButton() {
         content={
           isAnalysis
             ? "Copy a link with the current simulation parameters"
-            : "You can only share the analysis page"
+            : "You can only share when you had a simulation done"
         }
       >
         <Button shade="dark" disabled={!isAnalysis}>
--- a/apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
+++ b/apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
@@ -1,5 +1,4 @@
 import { Share1Icon } from "@radix-ui/react-icons";
-import { usePathname } from "next/navigation";
 import { useState } from "react";
 
 import Button from "#/components/Button";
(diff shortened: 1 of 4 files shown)


luizakp (dismissed):
Just tested here, working fine 👏

apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -159,9 +158,9 @@ export function PoolSimulatorProvider({ children }: PropsWithChildren) {
     setData(importedData);
   }
 
-  // useEffect(() => {
-  // if (pathname === "/poolsimulator/analysis") push(generateURL());
-  // }, [pathname]);
+  useEffect(() => {
+    if (isAnalysis) push(generateURL());
+  }, [customData, initialData]);
```
ribeirojose, line 163: 🤔 are we always adding this to the URL? Doesn't it impact performance rn? If it does I'd suggest we stop doing this and instead only leave the Share button working

apps/balancer-tools/src/components/Header.tsx
```diff
@@ -27,12 +28,14 @@ export function Header({
...
 }: IHeader) {
   return (
     <div className="flex h-20 w-full items-center border-b border-b-blue3 bg-blue2 p-4 text-white">
       <div className="mr-auto flex flex-1 justify-start">
         <Link
           href={linkUrl}
+          onClick={onLinkClick}
           className="flex items-center gap-3 justify-self-start"
         >
           {imageSrc && <Image src={imageSrc} height={50} width={50} alt="" />}
```
ribeirojose, line 41: Think we should probably make this composable. wdyt @luizakp ? We'd have something like `HeaderLink` and use it in children or pass it as a prop. IMO the best approach is to take both this link and the EndButton away and just have children to have a clearer API 😅

apps/balancer-tools/src/app/poolsimulator/(components)/ShareButton.tsx
```diff
@@ -8,23 +7,22 @@ import { Tooltip } from "#/components/Tooltip";
...
   return (
     <>
       <Tooltip
         content={
-          onInitialPage
-            ? "You can only share the analysis page"
-            : "Copy a link with the current simulation parameters"
+          isAnalysis
+            ? "Copy a link with the current simulation parameters"
+            : "You can only share the analysis page"
```
ribeirojose, line 19: Do people understand what the analysis page is? Not sure about our messaging around it in the app 😬

ribeirojose (approved):
lgtm -- left only a few remarks

PR #217, opened 2023-08-17
[APR] Historical APR chart for pool

Rawallon:
A simple line graph displaying historical APR. Currently, it retrieves the APR for each round through a loop, makes requests, and then plots the data on the graph from left to right, showing the progression from older to more recent data. Rounds in which a specific pool didn't exist are excluded. Additionally, it highlights the selected round with a marker.

image


apps/balancer-tools/src/lib/trimTrailingValues.test.ts
```diff
@@ -0,0 +1,71 @@
+import { trimTrailingValues } from "./utils";
```
ribeirojose, line 1: isn't this just a filter? 👀 do we expect to change it's functionality and want that change to propagate everywhere?

ribeirojose, line 1: 🤔 can you remind me why we needed this in the first place?

apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/round/[roundId]/page.tsx
```diff
@@ -19,6 +20,15 @@ export default async function Page({
       >
         <PoolOverviewCards roundId={roundId} poolId={poolId} />
       </Suspense>
+      <Suspense
+        fallback={
+          <div className="flex h-full w-full flex-col justify-center rounded-3xl">
+            Loading Chart...
+          </div>
```
Comment (ribeirojose) [inline apps/balancer-tools/src/app/apr/pool/[network]/[poolId]/round/[roundId]/page.tsx:27]: I think Lu created a standard Loading for this

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRPlot.tsx
```diff
@@ -0,0 +1,26 @@
+"use client";
```
ribeirojose, line 1: 👌  great use of the "use client" directive

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export default async function HistoricalAPRChart({
+  roundId,
+  poolId,
+}: {
+  roundId: string;
+  poolId: string;
+}) {
+  const HOVERTEMPLATE = "%{x}<br />%{y:.2f}% APR<extra></extra>";
+  const LAST_ROUND_ID = parseInt(Round.getAllRounds()[0].value);
```
ribeirojose, line 17: would move it out of the component if it's not dependent on the local logic here

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const APRPerRoundCords: { x: string[]; y: number[] } = { x: [], y: [] };
+  for (let index = 1; index < LAST_ROUND_ID; index++) {
+    const { apr } = await calculateAPRForPool({ poolId, roundId: index });
```
ribeirojose, line 20: Why not use a `Promise.all` approach? Would it be too many requests at once? If you need to throttle them you could use https://www.npmjs.com/package/p-throttle

ribeirojose (approved):
LGTM! But [this](https://github.com/bleu-studio/balancer-tools/pull/217/files#r1298482050) is a non-blocking #must imo

apps/balancer-tools/src/lib/trimTrailingValues.test.ts
Rawallon: It takes two arrays of numbers and a specified value to remove, and it returns new arrays with trailing occurrences of the specified value removed from both arrays. In that case I use to remove rounds where the pool didnt exist.

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRPlot.tsx
Rawallon: Thank you! I was somewhat unsure about it because I hadn't seen a component split like this yet.

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+    APRPerRoundCords.y.reverse(),
+    0,
+  );
+  const APRPerRoundData = {
+    name: "APR %",
+    hovertemplate: HOVERTEMPLATE,
+    x: trimmedArray.trimmedIn,
+    y: trimmedArray.trimmedOut,
+    line: { shape: "spline" },
+    type: "scatter",
```
luizakp, line 35:
```suggestion
    line: { shape: "spline" } as const,
    type: "scatter" as PlotType,
```

Without this you get type error

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
luizakp: yeah, here it is taking 1min20s to load everything ☠️

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  const chosenRoundMarkerIDX = APRPerRoundData.x.indexOf(`Round ${roundId}`);
```
luizakp, line 38:
```suggestion
  const chosenRoundMarkerIDX = APRPerRoundData.x.findIndex(item => item === `Round ${roundId}`);
```

because `APRPerRoundData.x` is `number[] | string[]`, therefore `indexOf()` expects type never Array<T>.indexOf(searchElement: never, fromIndex: number | undefined): number

apps/balancer-tools/src/lib/trimTrailingValues.test.ts
luizakp: on Pool Simulator we use this to remove the zones where the swap is not possible

PR #212, opened 2023-08-16
[APR] Pool listing as table

Rawallon:
I encountered some difficulties while attempting to style the table in accordance with the design specified in the Linear issue ([BAL-637](https://linear.app/bleu-llc/issue/BAL-637/pool-list-as-table)). Currently, I've reached a point where I'm considering discontinuing further styling efforts. The process of styling `<table>` and its children feels somewhat cumbersome due to certain properties not applying as expected, which is not the case with `<div>`.
I am contemplating reverting back to the default style from Table.tsx since I personally don't like the appearance of this table in the current PR.

image

PR #218, opened 2023-08-17
[APR] Refactor Pool List as table

luizakp:
This PR is a refactor of #212 

main changes:
- use of `calculatePoolStats`
- use Spinner as Suspense fallback
- transform PoolListTable into a client component

This PR also addresses @ribeirojose comments on #212


apps/balancer-tools/src/components/Table.tsx
```diff
@@ -1,14 +1,25 @@
...
+import { createContext, useContext } from "react";
+
+const TableContext = createContext({});
 
 const predefinedClasses = {
   gray: {
     solid: {
-      dark: { style: "bg-slate3", border: "border border-slate7 rounded" },
+      dark: { style: "bg-slate3", border: "border-0" },
+      darkWithBorder: {
```
ribeirojose, line 10: Why not simply add border to all?

ribeirojose (approved):
🎉

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
```diff
@@ -1,45 +1,75 @@
-import { ChevronDownIcon } from "@radix-ui/react-icons";
```
Rawallon, line 1: Good call on this, that's the only way to make it feel really dynamic and if the client's IP get rate limited he should eventually make this request

apps/balancer-tools/src/components/Table.tsx
luizakp: because on Internal Manager we use a blue table with no border 🤔

apps/balancer-tools/src/components/Table.tsx
```diff
@@ -1,14 +1,25 @@
 import cn from "clsx";
```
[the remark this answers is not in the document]
Rawallon: +1 for this since `createContext` needs it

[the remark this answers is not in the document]
luizakp: we didn't get suspense to work as we wanted on this particular case so we will try to look at this again later as this is not a `#must`

Rawallon (approved):
Looking good

PR #219, opened 2023-08-18
[APR] Pool TVL for a especific round

Rawallon:
This PR is similar to what's in balancer-alerts. It queries the subgraph to obtain the block number at the end of a round, which is then used as a parameter when querying for TVL. The Gnosis functionality is implemented within a separate function, there also should be an way to retrieve blocks for Optimism and Goerli. 
One point I'm unsure is that if no specific block is passed, it fetches the latest block value.


apps/balancer-tools/src/app/apr/(utils)/balancerAPI.ts
```diff
@@ -7,18 +7,32 @@ const DEFAULT_HEADERS: Record<string, string> = {
...
     }`;
 
-  public static async getPoolTotalLiquidityUSD(_chainId = 1, poolId: string) {
+  public static async getPoolTotalLiquidityUSD(
+    _chainId = 1,
+    poolId: string,
+    blockNumber: number,
+  ) {
     const { data } = await gql(
       BALANCER_GRAPHQL_API_URL,
```
ribeirojose, line 29: Just remember you'll also need to query the correct subgraph based on the chain that pool is in.

ribeirojose (changes requested):
This is looking good -- but the pattern we've followed is to create a query under /gql and call the generator. That way you'll have access to the hooks and fetcher directly in all chains that we support. I can show you around there but @luizakp and @devjoaov are also familiar with it FYI

luizakp:
> This is looking good -- but the pattern we've followed is to create a query under /gql and call the generator. That way you'll have access to the hooks and fetcher directly in all chains that we support. I can show you around there but @luizakp and @devjoaov are also familiar with it FYI

It looks like gql generated functions are all client side 🤔

luizakp:
> you can ask @mendesfabio for the blocks endpoints -- but did you search for them anywhere? You can find them in somewhere in Balancer's frontend config as well

They are empty on Balancer's frontend @ribeirojose. Created a [task for this](https://linear.app/bleu-llc/issue/BAL-659/add-sepolia-gnosis-and-optimism-subgraph-on-subgraphchainsblocks)

[Sepolia](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/sepolia/index.ts#L37)
[Optimism](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/optimism/index.ts#L36)
[Gnosis](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/gnosis-chain/index.ts#L39)

ribeirojose:
> > you can ask @mendesfabio for the blocks endpoints -- but did you search for them anywhere? You can find them in somewhere in Balancer's frontend config as well
> 
> They are empty on Balancer's frontend @ribeirojose. Created a [task for this](https://linear.app/bleu-llc/issue/BAL-659/add-sepolia-gnosis-and-optimism-subgraph-on-subgraphchainsblocks)
> 
> [Sepolia](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/sepolia/index.ts#L37) [Optimism](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/optimism/index.ts#L36) [Gnosis](https://github.com/balancer/frontend-v2/blob/46df6c2d6178236291196320447f6e217b500cd6/src/lib/config/gnosis-chain/index.ts#L39)

https://linear.app/bleu-llc/issue/BAL-659#comment-dadd94c2

PR #221, opened 2023-08-18
[Pool Simulator] Fix copy link not work on production

yvesfracari:


apps/balancer-tools/src/contexts/PoolSimulatorContext.tsx
```diff
@@ -167,7 +164,12 @@ export function PoolSimulatorProvider({ children }: PropsWithChildren) {
         setInitialData(state.initialData);
         setCustomData(state.customData);
         setIsAnalysis(true);
-        push("/poolsimulator");
+        const uri = window.location.toString();
+        window.history.replaceState(
+          {},
+          document.title,
+          uri.substring(0, uri.indexOf("#")),
+        );
```
ribeirojose, line 172: any reason not to call `push` with a hash param and use replaceState instead? Also, I think it's not the suggested approach to pass anything in the second param see https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

ribeirojose (dismissed):

PR #222, opened 2023-08-18
[Pool Simulator] Fix depth cost first render error

yvesfracari:


ribeirojose (approved):
👍 this is the kind of thing I'd be ok with merging right away and just letting us know on the slack channel

PR #226, opened 2023-08-19
[APR] Order pools by the biggest TVL on pool list

luizakp:
- Add network paths
- Add `getPoolList` to get PoolsByTotalLiquidity

https://www.loom.com/share/1586f0f01e7d4e9484e68774f0ba5750?sid=b20c26b9-ca90-4f57-b0c3-7f789f65a9a5

PR #227, opened 2023-08-19
[APR] Populate Top pools chart with real data

Rawallon:
Still takes upwards of a min to fetch it 💀

Here's a edited video:

https://github.com/bleu-studio/balancer-tools/assets/68453900/7a9f3e08-147c-421e-9293-4660e9bf5f23


apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
```diff
@@ -10,54 +8,60 @@ import {
...
+    const { apr } = await calculatePoolStats({
+      roundId,
+      poolId: gauge.pool.id,
+    });
+
+    if (apr && apr > 0.01) {
+      barCords.x.push(apr.toFixed(2));
+      barCords.y.push(gauge.pool.symbol);
+      barCords.colors.push(colors[index % colors.length]);
+    }
```
luizakp, line 46: Some strange behavior is happening here, you are a limit of 16 pools but only 13 are being rendered. Probably because not every pool from that 16 has APR. Might be something that #226  will help

apps/balancer-tools/src/app/apr/(utils)/sortChartData.ts
```diff
@@ -0,0 +1,34 @@
+interface iData {
```
ribeirojose, line 1: not a very descriptive name, would use `AxisData`. Also, not the standard we're using for interfaces (interface names should be TitleCase and we don't prefix with `I`) see the rest of the code.

apps/balancer-tools/src/app/apr/(utils)/sortChartData.ts
```diff
@@ -0,0 +1,34 @@
...
+  x: number[] | string[];
+  y: number[] | string[];
```
ribeirojose, line 3: would choose either one and convert before sending to this function

apps/balancer-tools/src/app/apr/(utils)/sortChartData.ts
ribeirojose: also consider whether you really need an interface here -- couldn't you just plug x and y as two function params? Why complicate things?

apps/balancer-tools/src/app/apr/(utils)/sortChartData.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Sorts data arrays by the values in the 'x' array.
+ *
+ * @param {Object} data - An object containing 'x' and 'y' arrays to be sorted.
+ * @returns {Array} An array containing sorted 'x' and corresponding 'y' arrays.
+ */
+export default function sortDataByX(data: iData) {
+  if (!Array.isArray(data.x) || !Array.isArray(data.y)) {
+    throw new Error("'x' and 'y' properties should be arrays.");
+  }
```
ribeirojose, line 15: curious about why this ended up being required -- didn't typechecking work?

luizakp (dismissed):
just wondering if p-throttle will help with this 🤔 my first take would be that plotly expects all the data (promise) to be resolved before trying to render

apps/balancer-tools/src/app/apr/(utils)/sortChartData.ts
```diff
@@ -0,0 +1,34 @@
...
+  y: number[] | string[];
+}
+
+/**
+ * Sorts data arrays by the values in the 'x' array.
+ *
+ * @param {Object} data - An object containing 'x' and 'y' arrays to be sorted.
+ * @returns {Array} An array containing sorted 'x' and corresponding 'y' arrays.
+ */
+export default function sortDataByX(data: iData) {
```
ribeirojose, line 12: Not really sure this function is needed. We just need to zip the array and order by one of two values right?
```suggestion
let x,y = [[1,2,3,...],[5,4,3....];
x.map((e, i) => [e, y[i]]).sort((x,y) => x[0]) # or y[0], whatever you want to do;
```

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
```diff
@@ -10,54 +8,60 @@ import {
...
+  for (let index = 0; index < votingGauges.length; index++) {
+    if (barCords.x.length >= POOL_QUANTITY_LIMIT) {
+      break;
+    }
+
+    const gauge = votingGauges[index];
+    const { apr } = await calculatePoolStats({
+      roundId,
+      poolId: gauge.pool.id,
+    });
```
ribeirojose, line 40: this will make iterating super slow

ribeirojose (changes requested):
`#must` fetch the data using promise.all

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
Rawallon: This logic will be altered on #226

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
Rawallon: Probably due to the `apr > 0.01` logic, it doesn't necessarily mean that it will display 16; it only displays a maximum of 16. However, I do think it's a valid point that displaying a fixed number is more user-friendly.

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
luizakp: now this will be handled on #235 right ?

luizakp (approved):
LGTM, the time waiting should be improved with #235

PR #228, opened 2023-08-19
[APR] p-throttle implementation

Rawallon:
Initial Implementation of Request Throttling.
In the first version of the request throttling mechanism, I've designed an instanced a Throttle within each component. This approach enables each component to have its own distinct throttle configurations. An alternative approach would involve creating the throttle instance at the page level and then passing it to the components as a prop. This second approach would centralize request throttling for all the requests originating from that particular page.

Here's in action with really batches of 3 and interval of 5s:

https://github.com/bleu-studio/balancer-tools/assets/68453900/07abc877-e465-4eb3-a6e4-1953d602b1a5

### DIFF
--- a/apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
+++ b/apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
@@ -9,6 +9,7 @@ import {
   TriangleUpIcon,
 } from "@radix-ui/react-icons";
 import { useRouter } from "next/navigation";
+import pThrottle, { ThrottledFunction } from "p-throttle";
 import { Dispatch, SetStateAction, useEffect, useState } from "react";
 
 import { Button } from "#/components";
--- a/apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
+++ b/apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
@@ -9,7 +9,7 @@ import {
   TriangleUpIcon,
 } from "@radix-ui/react-icons";
 import { useRouter } from "next/navigation";
-import pThrottle, { ThrottledFunction } from "p-throttle";
+import pThrottle from "p-throttle";
 import { Dispatch, SetStateAction, useEffect, useState } from "react";
 
 import { Button } from "#/components";
@@ -32,6 +32,11 @@ interface PoolTableData extends PoolStats {
   id: string;
   network: number;
 }
+
+export const throttle = pThrottle({
+  limit: 3,
+  interval: 1500,
+});
 
 export function PoolListTable({ roundId }: { roundId: string }) {
   const initialTableValues: PoolTableData[] = votingGauges
@@ -76,11 +81,6 @@ export function PoolListTable({ roundId }: { roundId: string }) {
     setOrder(sortOrder);
     handleSorting(accessor, sortOrder);
   };
-
-  const throttle = pThrottle({
-    limit: 3,
-    interval: 1500,
-  });
 
   return (
     <div className="flex w-full flex-1 justify-center text-white">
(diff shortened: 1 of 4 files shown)


apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
```diff
@@ -145,20 +153,27 @@ function TableRow({
...
   tableData,
   setTableData,
+  throttleHandler,
 }: {
   poolId: string;
   roundId: string;
   network: number;
   tableData: PoolTableData[];
   setTableData: Dispatch<SetStateAction<PoolTableData[]>>;
+  throttleHandler: ThrottledFunction<Argument, ReturnValue>;
```
Rawallon, line 163: Need help typing this

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
```diff
@@ -75,6 +76,12 @@ export function PoolListTable({ roundId }: { roundId: string }) {
     setOrder(sortOrder);
     handleSorting(accessor, sortOrder);
   };
+
+  const throttle = pThrottle({
+    limit: 3,
+    interval: 1500,
```
Rawallon, line 82: Arbritary numbers

ribeirojose, line 82:
```suggestion
    limit: 10,
    interval: 400,
```

package.json
```diff
@@ -48,5 +48,8 @@
     "overrides": {
       "ipfs-http-client": "^60"
     }
+  },
+  "dependencies": {
+    "p-throttle": "^5.1.0"
```
ribeirojose, line 53: ❌ this is dependency of `apps/balancer-tools`, not the entire repo

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
```diff
@@ -34,18 +46,14 @@ export default async function TopPoolsChart({ roundId }: { roundId: string }) {
     }
 
     const gauge = votingGauges[index];
-    const { apr } = await calculatePoolStats({
-      roundId,
-      poolId: gauge.pool.id,
-    });
+    const { apr } = await throttledCalculatePoolStats(roundId, gauge.pool.id);
```
ribeirojose, line 49: this is still iterating on a list and blocking rendering until we have every data we need, the only difference is that the requests are throttled. What value is this adding? Have you reflected about why this was needed here?

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
ribeirojose: why pass this as a prop?

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
Rawallon: The problem we're facing is that we're querying a bunch of pools. A direct implementation using Promise.all isn't suitable, as it would spam the server with simultaneous requests, resulting in immediate rate limiting. Consequently, these requests would persistently attempt retries, leading to subsequent requests slowing down. By using p-throttle, our intention is to navigate around the API's rate limitations. This approach should ideally enable us to accelerate the requests responses.

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
Rawallon: The `Table` instance includes `p-Throttle`, then we use `TableRow` to display data. This way enables `TableRow` to perform multiple queries without any constraints while the responsibility of limiting the queries is managed by `p-throttle` instanced on `Table`.

luizakp:
> > Initial Implementation of Request Throttling. In the first version of the request throttling mechanism, I've designed an instanced a Throttle within each component. This approach enables each component to have its own distinct throttle configurations. An alternative approach would involve creating the throttle instance at the page level and then passing it to the components as a prop. This second approach would centralize request throttling for all the requests originating from that particular page.
> > Here's in action with really batches of 3 and interval of 5s:
> > clideo_editor_55dc2c29e5e84705b9915c5d1d7b24d6.mp4
> 
> Just noticed there are two veBAL pools in the video 👀

this will be fixed on my PR #226

PR #229, opened 2023-08-21
[Pool Simulator] Add beta region limit on price impact curve

yvesfracari:
This PR:
- Adds the beta region on the price impact curve;
- Refactor Swap curve beta region functions;

image

As shown in the figure above, our beta region limits may not be accurate. This [issue](https://linear.app/bleu-llc/issue/BAL-663/beta-region-limit-calculation) will refer to this.


apps/balancer-tools/src/app/poolsimulator/(components)/Curves/ImpactCurve.tsx
```diff
@@ -186,6 +190,27 @@ export function ImpactCurve() {
     };
   };
 
+  const createBetaLimitsDataObject = (
+    x: number[],
+    y: number[],
+    legendGroup: string,
+  ) => {
+    // https://docs.xave.co/product-overview-1/fxpools/amm-faqs
```
ribeirojose, line 198: this link should be where in the place where we calculate these points instead of here right?

ribeirojose (approved):

PR #234, opened 2023-08-21
[APR] Average pool data if Round is not provided

Rawallon:
This PR does more than what says on the tin, it merges #235

### 235

- Adds API/Routes
- Alters `PoolListTable` to use API

### 234
- Refactor Pool page without roundID
- Refactor KPI component to allow components as content
- Created component for Pool Average APR/VTL
- Refactor logic for fetching pools on Top Pools Chart


apps/balancer-tools/src/app/apr/(components)/KPI.tsx
```diff
@@ -1,4 +1,10 @@
-export function KPI({ title, content }: { title: string; content: string }) {
+export function KPI({
+  title,
+  content,
+}: {
+  title: string;
+  content: string | JSX.Element;
```
ribeirojose, line 6: 👍

apps/balancer-tools/src/app/apr/(components)/OverviewCards.tsx
```diff
@@ -1,16 +1,25 @@
...
 interface CardDetail {
   title: string;
-  content: string;
+  content: JSX.Element | string;
 }
 
 function OverviewCards({ cardsDetails }: { cardsDetails: CardDetail[] }) {
   return (
     <div className="flex justify-between border border-blue6 bg-blue3 rounded p-4">
       <div className="flex flex-row flex-wrap xl:flex-nowrap gap-6 justify-evenly w-full">
```
ribeirojose, line 15: these cards should be the same size

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
```diff
@@ -145,20 +152,27 @@ function TableRow({
...
+  throttleHandler: typeof throttle;
 }) {
   const pool = new Pool(poolId);
   const [isLoading, setIsLoading] = useState(true);
 
   useEffect(() => {
     const fetchData = async () => {
+      const throttledFn = throttleHandler(
+        async (poolId: string, roundId: string) => {
+          return await fetcher(`/apr/api?poolid=${poolId}&roundid=${roundId}`);
```
ribeirojose, line 171: this call doesn't need to be throttled

Rawallon: I think it should be because they're going to be made alongside fetches from the `TopPoolsChart` component

apps/balancer-tools/src/app/apr/(components)/KPI.tsx
```diff
@@ -1,6 +1,12 @@
...
```
ribeirojose, line 6: I think `React.ReactNode` is the most accepting, then you wouldn't even need `string` here.

apps/balancer-tools/src/app/apr/(components)/OverviewCards.tsx
```diff
@@ -1,18 +1,25 @@
...
+    <div className="border border-blue6 bg-blue3 rounded p-4 gap-6 w-full grid auto-cols-fr	grid-flow-col">
+      {cardsDetails.map((cardDetail) => (
+        <Suspense
+          key={cardDetail.title}
+          fallback={<KPI title="" content={<Spinner size="sm" />} />}
+        >
           <KPI title={cardDetail.title} content={cardDetail.content} />
-        ))}
-      </div>
+        </Suspense>
```
ribeirojose, line 21: 👀 so perhaps the suspense boundary should be within the KPI card?

apps/balancer-tools/src/app/apr/(utils)/getHistoricalDataForPool.ts
```diff
@@ -0,0 +1,47 @@
...
+    Array.from({ length: endRoundId - startRoundId }, (_, index) =>
+      calculateAndPushStats(poolId, index + startRoundId),
+    ),
+  );
+
+export const generatePromisesForPoolList = async (
+  poolList: string[],
+  roundId: number = parseInt(Round.getAllRounds()[0].value),
+): Promise<Array<PoolStatsData>> =>
+  Promise.all(poolList.map((poolId) => calculateAndPushStats(poolId, roundId)));
```
ribeirojose, line 47: maybe I'm missing something still but not sure why we need this approach anymore 👀 we could just have our endpoint return whatever we need, no?

apps/balancer-tools/src/app/apr/round/(components)/TopPoolsChart.tsx
```diff
@@ -10,11 +10,29 @@ import { PlotType } from "plotly.js";
...
+        parseFloat(data.x[b] as string) - parseFloat(data.x[a] as string),
+    )
+    .reduce(
+      (sortedArrays, index) => {
+        sortedArrays[0].push(data.x[index]);
+        sortedArrays[1].push(data.y[index]);
+        return sortedArrays;
+      },
+      [[], []] as [(number | string)[], (number | string)[]],
+    );
```
ribeirojose, line 32: this should come sorted from the API and we shouldn't need to sort it here

apps/balancer-tools/src/app/apr/round/(components)/PoolListTable.tsx
ribeirojose: but if they're being handed to our own internal API that is going to cache these values that doesn't make a lot of sense -- also we're only showing 10 per page, so not a huge amount of requests to be doing at the same time.

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -0,0 +1,69 @@
+import { NextRequest, NextResponse } from "next/server";
+
+import votingGauges from "#/data/voting-gauges.json";
+
+import { calculatePoolStats } from "../(utils)/calculatePoolStats";
+import { Round } from "../(utils)/rounds";
+
+export async function GET(request: NextRequest) {
+  const poolId = request.nextUrl.searchParams.get("poolid");
+  const roundId = request.nextUrl.searchParams.get("roundid");
```
ribeirojose, line 10: should follow consistent casing that we're using elsewhere -- just do camelCase params here

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (!roundId && !poolId) {
+    return NextResponse.json({ error: "no round or poolId provided" });
+  }
```
ribeirojose, line 13: so we should return the default parameters -- the aggregated average values for each pool

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  if (roundId && poolId) {
+    const poolStats = await calculatePoolStats({ roundId, poolId });
+
+    return NextResponse.json(poolStats);
```
ribeirojose, line 18: ok, so apparently this is already [cached by default](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#caching) 👍

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -0,0 +1,69 @@
...
+      votingShare: averagedValues.votingShare / numResults,
+    };
+
+    const perRound = results.map((result, index) => {
+      return {
+        [parseInt(roundGaugeAdded.value) + index]: result,
+      };
+    });
+
+    return NextResponse.json({ perRound, average: averageResult });
```
ribeirojose, line 67: 👍 looks good to me -- we're just missing sorting/filtering here, right?

ribeirojose (dismissed):
`#must` have sort/filter/limit in the API to avoid doing that in client components and to leverage caching there -- but we can incorporate that to another PR

apps/balancer-tools/src/app/apr/(components)/KPI.tsx
Rawallon: Actually had forgotten about `React.ReactNode`

apps/balancer-tools/src/lib/balancer/gauges.ts
```diff
@@ -3,7 +3,7 @@ import { NetworkChainId } from "@bleu-balancer-tools/utils";
 import POOLS_WITH_GAUGES from "#/data/voting-gauges.json";
 
 export const POOLS_WITH_LIVE_GAUGES = POOLS_WITH_GAUGES.filter(
-  (pool) => !pool.gauge.isKilled && pool.gauge.addedTimestamp,
```
ribeirojose: pools whose gauge doesn't have a timestamp have a different type of gauge and shouldn't be accounted for

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -89,10 +89,7 @@ const fetchDataForPoolId = async (
   const gaugesData = await Promise.allSettled(promises);
 
   const resolvedPoolData = gaugesData
-    .filter(
-      (result): result is PromiseFulfilledResult<PoolStatsResults> =>
-        result.status === "fulfilled",
-    )
+    .filter(result.status === "fulfilled")
```
ribeirojose, line 92: ?

PR #235, opened 2023-08-22
[APR] move logic from client side to api

Rawallon:


Rawallon:
Sent commits from this branch on #234 [here](https://github.com/bleu-studio/balancer-tools/pull/234#commits-pushed-8bf8b38) since we decided to close #226

PR #237, opened 2023-08-22
[APR] "Load more" button on pool table

Rawallon:
Fetches and display more gauges on clicking "load more" on pool table

https://github.com/bleu-studio/balancer-tools/assets/68453900/c16d39a5-bf08-419c-bba7-b571a29771bd


Rawallon:
This logic is going to change on #238

PR #239, opened 2023-08-23
Download voting gauges from api on app build

ribeirojose:


apps/balancer-tools/src/lib/balancer/gauges.ts
```diff
@@ -73,18 +84,16 @@ export class Gauge {
     }
 
     const data = GAUGE_DATA.find(
-      (g) => g.address.toLowerCase() === address.toLowerCase(),
+      (g) => g.gauge.address.toLowerCase() === address.toLowerCase(),
```
Rawallon, line 87: This is breaking, I think it should be
```suggestion
      (g) => g.address.toLowerCase() === address.toLowerCase(),
```

ribeirojose: hm need to take a look into why -- but the new structure from the API is actually slightly different -- the parent object is a pool, with a gauge nested within

PR #241, opened 2023-08-25
[APR] Better TVL calculation

Rawallon:
Insted of getting the TVL value from the subgraph, which could contain wrong value for coins we now get the amount for which coin then get its value for a given round and calculte it on our end


packages/gql/src/pool-snapshots/snapshots.ts
```diff
@@ -0,0 +1,12 @@
+import { gql } from "graphql-tag";
+
+export const PoolSnapshots = gql`
+  query PoolSnapshot($pool: String, $timestamp: Int) {
+    poolSnapshots(where: { pool: $pool, timestamp_gte: $timestamp }) {
```
ribeirojose, line 5: would also need to be sorted by timestamp and/or return the timestamp in the response, no?

packages/gql/codegen.ts
```diff
@@ -111,6 +112,27 @@ export const SUBGRAPHS = {
       return this.endpoints()[network];
     },
   },
+  [Subgraph.PoolSnapshots]: {
```
ribeirojose, line 115: we already have this subgraph here, so you don't need to add it again. You can just add a new query to the existing balancer subgraph

packages/gql/src/pool-snapshots/snapshots.ts
```diff
@@ -0,0 +1,12 @@
...
+      pool {
+        id
+      }
+      amounts
```
ribeirojose, line 9: why not get the total liquidity here? The instantaneous subgraph value might be off but the point of the snapshot is that that value is usually gonna be correct. See [this quick query I ran](https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2/graphql?query=query+%7B%0A++poolSnapshots(where:+%7Bpool:%220x87a867f5d240a782d43d90b6b06dea470f3f8f22000200000000000000000516%22,+timestamp_gt:+1692884610%7D,+first:1)+%7B%0A++++liquidity%0A++++pool+%7B%0A++++++id%0A++++%7D%0A++%7D%0A%7D): 

```graphql
query {
  poolSnapshots(where: {pool:"0x87a867f5d240a782d43d90b6b06dea470f3f8f22000200000000000000000516", timestamp_gt: 1692884610}, first:1) {
    liquidity
    pool {
      id
    }
  }
}
```
result:
```json
{
  "data": {
    "poolSnapshots": [
      {
        "liquidity": "17571798.52955511399359994811209584",
        "pool": {
          "id": "0x87a867f5d240a782d43d90b6b06dea470f3f8f22000200000000000000000516"
        }
      }
    ]
  }
}
```

Rawallon: Because it yesterday it was returning the same wrong value as the query we were running before

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -28,6 +30,21 @@ const getDataFromCacheOrCompute = async <T>(
...
+  );
+
+  const coinPrice = await getPoolTokenPriceByRound(poolId, round);
+
+  return tokenAmount.reduce(
+    (sum, value1, index) =>
+      sum + parseFloat(value1) * parseFloat(coinPrice[index]),
+    0,
+  );
+};
```
ribeirojose, line 46: I think this makes sense for us to get more precise with current round -- but as mentioned below and in [a Slack message](https://bleu-llc.slack.com/archives/C05FE3PLPP0/p1692971277559029) I sent I think it's going to require less API calls and be more consistent with Balancer's own FE to get the historical data from the snapshot

packages/gql/src/pool-snapshots/snapshots.ts
ribeirojose: you mean in the pools subgraph, right? yesterday there was no snapshot for yesterday yet 😅

ribeirojose:
closed by https://github.com/bleu-studio/balancer-tools/commit/48d7e608016dddb764914e0bb43fb3a9a180b162

PR #245, opened 2023-08-26
[APR] Minor fixes on FE

Rawallon:


apps/balancer-tools/src/app/apr/layout.tsx
```diff
@@ -1,24 +1,28 @@
...
 
 import HeaderEndButton from "./(components)/HeaderEndButton";
 
-export default function Layout({ children }: React.PropsWithChildren) {
-  const { roundId, poolId, network } = useParams();
-  invariant(!Array.isArray(roundId), "roundId cannot be a list");
-  invariant(!Array.isArray(poolId), "poolId cannot be a list");
-  invariant(!Array.isArray(network), "network cannot be a list");
+export const metadata: Metadata = {
+  title: "Balancer - APR Enhancement Calculator",
```
Rawallon, line 9: @ribeirojose Could you validate this name?

apps/balancer-tools/src/app/apr/layout.tsx
```diff
@@ -1,24 +1,28 @@
...
+  children: React.ReactNode;
+  params: { roundId: string; poolId: string; network: string };
+}) {
+  const { roundId, poolId, network } = params;
   return (
     <div className="flex h-full flex-col">
       <Header
         linkUrl={"/apr"}
-        title={"Pool Simulator"}
+        title={"APR Enhancement Calculator"}
```
Rawallon, line 25: Here too

apps/balancer-tools/src/app/apr/layout.tsx
ribeirojose: We can call it Historical APR

apps/balancer-tools/src/app/apr/layout.tsx
ribeirojose: Would create a constant with the name and use it in both places

apps/balancer-tools/src/app/apr/(components)/HeaderEndButton.tsx
```diff
@@ -30,10 +30,9 @@ export default function HeaderEndButton({
     if (!poolId && (!roundId || roundId === "current")) {
       router.push(`/apr/round/${LAST_ROUND_ID}`);
       setSelectedRound(LAST_ROUND_ID);
-    } else {
-      setSelectedRound(undefined);
+    } else if (typeof roundId === "undefined") {
+      setSelectedRound("");
     }
```
ribeirojose, line 35: was this actually causing any issues?

ribeirojose (approved):
lgtm -- let's only call the app `Balancer Historical APRs`

apps/balancer-tools/src/app/apr/(components)/HeaderEndButton.tsx
Rawallon: It just be like that, Should be fixed by now by [this pr](https://github.com/radix-ui/primitives/issues/1569#issuecomment-1568518196) but that didn't work, so I did the suggested workaround1

ribeirojose (approved):

PR #252, opened 2023-08-30
[APR] Refactor API tor return different APR types

luizakp:
this PR refactors the API return on APR

previously: 
```
apr: number
```

now: 
```
  apr: {
    total: number;
    breakdown: {
      veBAL: number;
    };
  };
```

- for round views I'm considering APR total, which rn is the same as veBAL APR

### DIFF
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -94,12 +94,6 @@ export async function calculatePoolStats({
   ]);
 
   const apr = calculateRoundAPR(round, votingShare, tvl, balPriceUSD);
-  const aprBefore =
-    balPriceUSD && tvl && votingShare
-      ? calculateRoundAPRBefore(round, votingShare, tvl, balPriceUSD) * 100
-      : -1;
-
-  console.log({ apr, aprBefore });
 
   return {
     roundId: Number(roundId),
@@ -131,14 +125,4 @@ function calculateRoundAPR(
       veBAL: vebalAPR,
     },
   };
-}
-
-function calculateRoundAPRBefore(
-  round: Round,
-  votingShare: number,
-  tvl: number,
-  balPriceUSD: number,
-): number {
-  const emissions = balEmissions.weekly(round.endDate.getTime() / 1000);
-  return (WEEKS_IN_YEAR * (emissions * votingShare * balPriceUSD)) / tvl;
 }
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -93,10 +93,13 @@ export async function calculatePoolStats({
     ),
   ]);
 
-  const apr =
+  const apr = calculateRoundAPR(round, votingShare, tvl, balPriceUSD);
+  const aprBefore =
     balPriceUSD && tvl && votingShare
-      ? calculateRoundAPR(round, votingShare, tvl, balPriceUSD) * 100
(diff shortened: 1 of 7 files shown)


apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -115,7 +112,17 @@ function calculateRoundAPR(
...
   balPriceUSD: number,
-): number {
+) {
   const emissions = balEmissions.weekly(round.endDate.getTime() / 1000);
-  return (WEEKS_IN_YEAR * (emissions * votingShare * balPriceUSD)) / tvl;
+  const vebalAPR =
+    balPriceUSD && tvl && votingShare
+      ? ((WEEKS_IN_YEAR * (emissions * votingShare * balPriceUSD)) / tvl) * 100
+      : -1;
+
```
ribeirojose, line 121: emit a warning if this result is -1, preferably we'd be warned on Sentry when this happens with the data to reproduce it.

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -307,10 +329,14 @@ function sortAndLimit(
...
     .sort((aValue, bValue) => {
-      const valueA = aValue[sortProperty];
-      const valueB = bValue[sortProperty];
+      let valueA = aValue[sortProperty];
+      let valueB = bValue[sortProperty];
+
+      if (sortProperty === "apr") {
+        valueA = (valueA as (typeof aValue)["apr"]).total;
+        valueB = (valueB as (typeof bValue)["apr"]).total;
+      }
```
ribeirojose, line 338: 👍

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
```diff
@@ -26,23 +26,23 @@ export default async function HistoricalAPRChart({
   const aprPerRoundCords = Object.entries(results.perRound).reduce(
     (cords, [_, result]) => {
       cords.x.push(getRoundName(result.roundId));
-      cords.y.push(result.apr);
+      cords.y.push(result.apr.breakdown.veBAL);
```
ribeirojose, line 29: could we also add total and veBAL as separate series in this plot? This would already prepare us for when the total considers other APR sources as well.

ribeirojose (approved):
👌 only left a minor suggestion

Rawallon (approved):
Lets get this merging!

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalAPRChart.tsx
luizakp: sure! I was going to add this only when we had the fees APR so we don't confuse the potential users

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -93,10 +95,14 @@ export async function calculatePoolStats({
...
-      ? calculateRoundAPR(round, votingShare, tvl, balPriceUSD) * 100
-      : -1;
+  const apr = calculateRoundAPR(round, votingShare, tvl, balPriceUSD);
+
+  if (apr.total === -1 || apr.breakdown.veBAL === -1) {
+    Sentry.captureMessage("vebalAPR resulted in -1", {
+      level: "warning",
+      extra: { balPriceUSD, tvl, votingShare, roundId, poolId, apr },
+    });
+  }
```
luizakp, line 105: @ribeirojose sentry warning added here so we can know which pool and on which round the APR wasn't calculated

ribeirojose (dismissed):

PR #254, opened 2023-08-31
[APR] More filters component

Rawallon:
https://www.loom.com/share/113aa26f8b414273a9e6cb1fae9e28e9


Rawallon:
https://www.loom.com/share/113aa26f8b414273a9e6cb1fae9e28e9

luizakp:
Is there a task for clear filters? If so please link it here =)

luizakp:
we have a OP pool 🤔 
image

luizakp:
> Is there a task for clear filters? If so please link it here =)

https://www.loom.com/share/08679fd43ebb4f0c977ea48dbed862b4?sid=1daadc36-6c68-4a43-bd2d-a567af92972f

luizakp:
Also here on the UI the "more filters" has more y-padding than the input

image

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,274 @@
...
+import { PoolTypeEnum } from "../../(utils)/calculatePoolStats";
+import { INITIAL_MIN_TVL } from "../../(utils)/getFilteredApiUrl";
+
+const networksOnBalancer = {
+  "1": Network.Ethereum,
+  "137": Network.Polygon,
+  "42161": Network.Arbitrum,
+  "100": Network.Gnosis,
+  "1101": Network.PolygonZKEVM,
+};
```
luizakp, line 28: This already exists on utils `index.ts` const `networkIdEnumMap`

Rawallon: Yeah, I pretty much copied that, but I removed the networks that aren't available on Balancer's dropdown. Would you prefer me to use networkIdEnumMap's?

luizakp:
nice video!

Rawallon:
> we have a OP pool 🤔 if that's correct I would add the OP symbol too image

Kinda the same question. Should we display it even though it isn't on Balancer's FE?

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
luizakp: you can use the `networkIdEnumMap` and do something like this:

```
const filteredKeys = [
  String(NetworkChainId.ETHEREUM),
  String(NetworkChainId.POLYGON),
  String(NetworkChainId.ARBITRUM),
  String(NetworkChainId.GNOSIS),
  String(NetworkChainId.POLYGONZKEVM),
];

const networksOnBalancerTest = Object.fromEntries(
  Object.entries(networkIdEnumMap).filter(([key]) => filteredKeys.includes(key))
);
```

and this is the same as the object you created:

```
{
  networksOnBalancer: {
    '1': 'ethereum',
    '100': 'gnosis',
    '137': 'polygon',
    '1101': 'polygon-zkevm',
    '42161': 'arbitrum'
  },
  networksOnBalancerTest: {
    '1': 'ethereum',
    '100': 'gnosis',
    '137': 'polygon',
    '1101': 'polygon-zkevm',
    '42161': 'arbitrum'
  }
}
```

luizakp:
> > we have a OP pool 🤔 if that's correct I would add the OP symbol too image
> 
> Kinda the same question. Should we display it even though it isn't on Balancer's FE?

It's on Balancer's FE actually. So we should display it
image

Rawallon:
Also adjusted Input's height

image

apps/balancer-tools/src/app/apr/(components)/MultiSelectDropdown.tsx
```diff
@@ -111,7 +113,7 @@ export const MultiSelectDropdown = ({
                         toggleMenu();
                       }}
                       onKeyDown={handleOnKeyDown}
-                      className="hover:shadow-[0] focus:shadow-[0] shadow-none flex-1"
+                      className="hover:shadow-[0] focus:shadow-[0] shadow-none flex-1 !h-6"
```
ribeirojose, line 116: 🤔 why did we remove the shadows here? Also curious about why the height wasn't applying and needed a `!` here

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
```diff
@@ -4,6 +4,13 @@ import { SearchParams } from "../round/[roundId]/page";
 export const INITIAL_MIN_TVL = 1000;
 export const INITIAL_LIMIT = 10;
 
+interface ExpectedSearchParams extends SearchParams {
+  minTVL?: string | undefined;
```
ribeirojose, line 8:
```suggestion
  minTVL?: string | undefined;
```
is redundant. You can just use `prop?: string`: https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters

packages/utils/index.ts
```diff
@@ -118,6 +118,21 @@ export const networkIdEnumMap = {
   "1101": Network.PolygonZKEVM,
 };
 
+const filteredKeys = [
```
ribeirojose, line 121: what does this name mean? 😅

apps/balancer-tools/src/app/apr/(components)/MultiSelectDropdown.tsx
Rawallon: This component has the class `h-[35px]` was being applied, I added important to overwrite it

packages/utils/index.ts
```diff
@@ -118,6 +118,21 @@ export const networkIdEnumMap = {
...
+  String(NetworkChainId.ETHEREUM),
+  String(NetworkChainId.OPTIMISM),
+  String(NetworkChainId.GNOSIS),
+  String(NetworkChainId.POLYGON),
+  String(NetworkChainId.POLYGONZKEVM),
+  String(NetworkChainId.ARBITRUM),
+];
+
+export const networksOnBalancer = Object.fromEntries(
```
ribeirojose, line 130:
```suggestion
export const NETWORKS_ON_BALANCER = Object.fromEntries(
```

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+              onChange={(e: ChangeEvent<HTMLInputElement>) =>
+                onChange(`min${name}`, e.target.value)
+              }
+              value={selectedAttributes[`min${name}`] || ""}
+            />
+            <span>-</span>
+            <BaseInput
+              className="flex-1"
+              name={`max${name}`}
+              placeholder="Maximum"
```
ribeirojose, line 255:
```suggestion
              placeholder="max"
```

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+  return (
+    <AccordionItem value={String(name)}>
+      <AccordionTrigger>{label}</AccordionTrigger>
+      <AccordionContent>
+        <div className="py-2 flex flex-col">
+          <div className="flex items-center gap-1">
+            <BaseInput
+              className="flex-1"
+              name={`min${name}`}
+              placeholder="Minimum"
```
ribeirojose, line 245:
```suggestion
              placeholder="min"
```

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+              onChange={(e: ChangeEvent<HTMLInputElement>) =>
+                onChange(`min${name}`, e.target.value)
```
ribeirojose, line 247: `e.target.name` is probably the same as `min${name}`, no? then you could also change your onChange fn

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+                              label={option}
+                            />
+                          </div>
+                        ))}
+                      </div>
+                    </AccordionContent>
+                  </AccordionItem>
+                ))}
+                <AccordionItemMinMax
+                  name="Tvl"
```
ribeirojose, line 199: why use name in TitleCase here when everywhere else they're camelCase?

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+    });
+  };
+
+  useEffect(() => {
+    const timer = setTimeout(pushQueryparamsURL, 700);
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [selectedAttributes]);
+
```
ribeirojose, line 141: cool! You could also use the `useDebounce` hook we've defined

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const countValues = useCallback(
```
ribeirojose, line 142:
```suggestion
  const countNonNullValues = useCallback(
```

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
```diff
@@ -0,0 +1,266 @@
...
+        updatedAttributes[value] = searchParams.get(value)?.split(",") || [];
+      }
+    });
+    setSelectedAttributes(updatedAttributes);
+  }, []);
+
+  const pushQueryparamsURL = () => {
+    const current = new URLSearchParams(Array.from(searchParams.entries()));
+
+    Object.entries(selectedAttributes).map(([name, values]) => {
```
ribeirojose, line 89: would make sure this is wrapped in a useCallback with `selectedAttributes` in the dep array

ribeirojose (approved):

ribeirojose:
🔝

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
Rawallon: This is going to concatenated with "min/max{Name}"

apps/balancer-tools/src/app/apr/round/(components)/MoreFiltersButton.tsx
Rawallon: I tried it and it does'nt work in for this use case, react throws an exception `Error: Invalid hook call. Hooks can only be called inside of the body of a function component.`, because of its useState

PR #256, opened 2023-09-04
[APR] expose collectedFeesUSD to API

luizakp:
This PR just exposes feeDiff as collectedFeesUSD to be shown on future chart on https://linear.app/bleu-llc/issue/BAL-712/add-tabs-with-historical-volume-tvl-collected-fees

Example of possible chart: 
image


ribeirojose (approved):
🔝

PR #257, opened 2023-09-04
[APR] Tokens on pool page

Rawallon:
[Deployed](https://tools-monorepo-jbi2533os-ribeirojose.vercel.app/apr/pool/ethereum/0x6f0ed6f346007563d3266de350d174a831bde0ca0001000000000000000005db/round/74)

<details><summary>Balancer's</summary>

image

</details>

# Weighted Pools

image

# No round provided

image

# Before Pool's creation

image

When the pools is not weighted it doesnt show the "Volume" col

### DIFF
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -116,24 +116,13 @@ export async function calculatePoolStats({
     ),
   ]);
 
-  let tokens;
-  try {
-    tokens = pool.tokens.map((token, idx) => ({
-      logoSrc: token.logoSrc,
-      address: token.address,
-      symbol: token.symbol,
-      weight: token.weight,
-      balance: tokenBalance[idx].balance,
-    }));
-  } catch (error) {
-    tokens = pool.tokens.map((token) => ({
-      logoSrc: token.logoSrc,
-      address: token.address,
-      symbol: token.symbol,
-      weight: token.weight,
-      balance: null,
-    }));
-  }
+  const tokens = pool.tokens.map((token, idx) => ({
+    logoSrc: token.logoSrc,
+    address: token.address,
+    symbol: token.symbol,
+    weight: token.weight,
+    balance: tokenBalance?.[idx]?.balance,
+  }));
 
   const apr = calculateRoundAPR(round, votingShare, tvl, balPriceUSD, feeAPR);
 
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -6,7 +6,7 @@ import { Pool } from "#/lib/balancer/gauges";
 import { pools } from "#/lib/gql/server";
 
 import { PoolStatsData, PoolTokens } from "../api/route";
-import { getBALPriceByRound } from "./getBALPriceByRound";
+import { getBALPriceByRound, getTokenPriceByRound } from "./getBALPriceByRound";
 import { getPoolRelativeWeight } from "./getRelativeWeight";
 import { Round } from "./rounds";
 
@@ -70,6 +70,37 @@ const fetchPoolTVLFromSnapshotAverageFromRange = async (
   ];
 };
 
+async function calculateTokensStats(
+  roundId: string,
+  poolTokenData: PoolTokens[],
+  poolNetwork: string,
+  tokenBalance: { symbol: string; balance: string }[],
+) {
+  const totalBalance = poolTokenData.reduce((acc, token, idx) => {
+    const balance = parseFloat(tokenBalance?.[idx]?.balance);
+    if (!isNaN(balance)) {
+      return acc + balance;
+    }
+    return acc;
+  }, 0);
+
+  const tokenPromises = poolTokenData.map(async (token, idx) => {
+    const tokenPrice = await getTokenPriceByRound(
+      Round.getRoundByNumber(roundId),
+      token.address,
+      parseInt(poolNetwork),
+    );
+    token.price = tokenPrice;
+    token.balance = tokenPrice * parseFloat(tokenBalance?.[idx]?.balance);
+    token.percentageValue =
+      ((tokenPrice * parseFloat(tokenBalance?.[idx]?.balance)) / totalBalance) *
+      100;
+    return token;
+  });
+
+  return Promise.all(tokenPromises);
+}
+
 export async function calculatePoolStats({
   roundId,
   poolId,
@@ -116,13 +147,12 @@ export async function calculatePoolStats({
     ),
   ]);
 
-  const tokens = pool.tokens.map((token, idx) => ({
-    logoSrc: token.logoSrc,
-    address: token.address,
-    symbol: token.symbol,
-    weight: token.weight,
-    balance: tokenBalance?.[idx]?.balance,
-  }));
+  const tokens = await calculateTokensStats(
+    roundId,
+    pool.tokens,
+    network,
+    tokenBalance,
+  );
 
   const apr = calculateRoundAPR(round, votingShare, tvl, balPriceUSD, feeAPR);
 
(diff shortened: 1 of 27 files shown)


apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+}: {
+  poolStats: PoolStatsData;
+}) {
+  return (
+    <>
+      <Table color="blue" shade={"darkWithBorder"}>
+        <Table.HeaderRow>
+          <Table.HeaderCell>Name</Table.HeaderCell>
+          <Table.HeaderCell classNames="w-1/6 text-end p-4">
+            Volume
```
luizakp, line 20:
```suggestion
            Balance
```

luizakp:
`#would` add `Token %` on the token table

Balancer example:

image

luizakp:
It would be nice to add the token balance value in USD too, but this can be another task if needed

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
Rawallon: I'm using balancer's nomenclature there

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+              <Table.BodyCellLink
+                href={`https://etherscan.io/address/${token.address}`}
+                tdClassNames="w-6"
+                linkClassNames="justify-end w-full"
+              >
+                {parseFloat(token.balance).toLocaleString("en-US", {
+                  minimumFractionDigits: 0,
+                  maximumFractionDigits: 0,
+                })}
+              </Table.BodyCellLink>
```
luizakp, line 73: They are flipped
image

luizakp:
`#must` add a title `Current Pool Composition`, otherwise a user might think that this composition during a selected round

Rawallon:
> It would be nice to add the token balance value in USD too, but this can be another task if needed

There should already be a value for specific round, or do you mean for a single token?  We could also add the avg value for that token, but I think it would add a lot of aditional requests

luizakp:
Handle when user select a round where the pool didn't exist

image

luizakp:
> > It would be nice to add the token balance value in USD too, but this can be another task if needed
> 
> There should already be a value for specific round, or do you mean for a single token? We could also add the avg value for that token, but I think it would add a lot of aditional requests

hmmm in that case I suggest to break it like Balancer does. One column is tha balance, which is the amount of token. Then, value is that balance in USD, which is balance * current token price

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
luizakp: can you provide an example? Maybe you are thinking about the "Volume" on the charts, but on the pool composition they use this nomenclature:

image

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokens.tsx
```diff
@@ -0,0 +1,36 @@
...
+    poolData.perRound.map((pool) => {
+      pool.tokens.map(async (token) => {
+        // Is there a way to type this without recreating the interface?
+        // @ts-ignore 2345
+        const tprice = await getTokenPriceByRound(
+          Round.getRoundByNumber(roundId),
+          token.address,
+          parseInt(pool.network),
+        );
+        // @ts-ignore 2345
```
luizakp, line 30: 👀

Rawallon: There's no way around it is there?

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -77,36 +81,59 @@ export async function calculatePoolStats({
...
+  ]);
+
+  let tokens;
+  try {
+    tokens = pool.tokens.map((token, idx) => ({
+      logo: token.logo,
+      address: token.address,
+      symbol: token.symbol,
+      weight: token.weight,
+      balance: tokenBalance[idx].balance,
```
ribeirojose, line 126:
```suggestion
      balance: tokenBalance?.[idx]?.balance,
```

apps/balancer-tools/src/app/apr/(utils)/getBALPriceByRound.ts
```diff
@@ -1,16 +1,30 @@
...
+    1,
+  );
+};
+
+export const getTokenPriceByRound = async (
+  round: Round,
+  tokenAddress: string,
+  tokenNetwork: number,
+) => {
+  const token = `${networkFor(tokenNetwork)}:${tokenAddress}`;
```
ribeirojose, line 20: this first part must be lowercase

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokens.tsx
```diff
@@ -0,0 +1,40 @@
...
+      const tokenPrice = await getTokenPriceByRound(
+        Round.getRoundByNumber(roundId),
+        token.address,
+        parseInt(poolData.perRound[0].network),
+      );
+      token.price = tokenPrice;
+      token.percentageValue =
+        ((tokenPrice * parseFloat(token.balance)) / totalBalance) * 100;
+    });
+  }
```
ribeirojose, line 38: any reason not to return this with the API response, or maybe throw this to a different endpoint?

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+export default function PoolTokensTable({
+  poolTokensStats,
+}: {
+  poolTokensStats: PoolTokens[];
+}) {
+  const tokenBalanceUSD = (balance: string) =>
+    parseFloat(balance).toLocaleString("en-US", {
+      minimumFractionDigits: 0,
+      maximumFractionDigits: 0,
+    });
```
ribeirojose, line 18: IMO we should return early if there's nothing in `poolTokensStats`

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+            <Table.HeaderCell classNames="w-1/6 text-end p-4">
+              Token %
+            </Table.HeaderCell>
+          )}
+        </Table.HeaderRow>
+        <Table.Body>
+          {poolTokensStats.map((token) => (
+            <Table.BodyRow classNames="hover:bg-blue4 hover:cursor-pointer duration-500">
+              <Table.BodyCellLink
+                href={`https://etherscan.io/address/${token.address}`}
```
ribeirojose, line 49: tokens won't all be in etherscan -- remember this is multichain. I think we have a canonical list of chain scanners somewhere

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+                </Table.BodyCellLink>
+              )}
+              {token.price && poolTokensStats[0].balance && (
+                <Table.BodyCellLink
+                  href={`https://etherscan.io/address/${token.address}`}
+                  tdClassNames="w-6"
+                  linkClassNames="justify-end w-full"
+                >
+                  {`$${(
+                    token.price! * parseFloat(token.balance)
```
ribeirojose, line 94: why use the non null assertion (!) here?

ribeirojose (approved):
lgtm only a few sharp edges left. Is this already mobile-friendly?

apps/balancer-tools/src/app/apr/pool/(components)/PoolTokensTable.tsx
Rawallon: There's a `if token.price && poolTokensStats[0].balance`, if that code gets reached should mean that it isn't undefined

PR #259, opened 2023-09-05
[APR] Minor UI fixes

Rawallon:
### Changes
 - Added table title
 - Altered table tooltip
 - If table start with <10 pools it will show "All pools have been loaded" instead of "load more"
 - Historical APR now should only show integer
 
### Fixes
 - Redirect with default sorting if not on searchParams
 - Load more not using selected filters
 - Tokens had the same URL

#### Table

Title Tooltip: All values are calculated at the end of the round

APR Tooltip: The value displayed is the min APR

image


luizakp (approved):
lgtm

PR #262, opened 2023-09-05
[APR] Mobile design revamp

Rawallon:
https://github.com/bleu-studio/balancer-tools/assets/68453900/0296bd4f-ec72-4d74-960d-bd5a51d4cc6c

### DIFF
--- a/apps/balancer-tools/src/app/apr/(components)/KPI.tsx
+++ b/apps/balancer-tools/src/app/apr/(components)/KPI.tsx
@@ -2,7 +2,7 @@ import { InfoCircledIcon } from "@radix-ui/react-icons";
 import { Suspense } from "react";
 
 import { Spinner } from "#/components/Spinner";
-import { Tooltip } from "#/components/Tooltip";
+import { TooltipMobile } from "#/components/TooltipMobile";
 
 function KPICardWrapper({ children }: { children: React.ReactNode }) {
   return (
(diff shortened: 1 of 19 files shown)


apps/balancer-tools/src/app/apr/(components)/KPI.tsx
```diff
@@ -2,7 +2,7 @@ import { InfoCircledIcon } from "@radix-ui/react-icons";
 import { Suspense } from "react";
 
 import { Spinner } from "#/components/Spinner";
-import { Tooltip } from "#/components/Tooltip";
+import { TooltipMobile } from "#/components/TooltipMobile";
```
ribeirojose, line 5: Do we need the name change? Couldn't this simply be a tooltip that works on mobile? 🤷

ribeirojose (dismissed):
This is really great stuff, thanks @Rawallon ! 🙇 🙇 🙇 Lots of seemingly small changes but the care you took with each of them really shines through!

IMG_F15FB555B6F8-1
think the centering / padding of the Network column is a bit off, perhaps because the token the same bg? The token names are not left aligned, so could as well be that. Or we could try with justifying the network icons centrally.

Rawallon:
> think the centering / padding of the Network column is a bit off, perhaps because the token the same bg? The token names are not left aligned, so could as well be that. Or we could try with justifying the network icons centrally.

Fixed it. It was centered in the middle since its <tr> default behaviour.
For future reference: Table's child element wont have width/height defined unless the Table also has. [Codepen demonstrating it](https://codepen.io/rawallon/pen/vYvybgx).

PR #265, opened 2023-09-11
[APR] use avg of vebal price instead of only one point on apr calc

Rawallon:
This shouldn't change anything majorly, just a refactor se we use average Bal price instead of one point at the end of the round

### DIFF
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -6,7 +6,7 @@ import { Pool } from "#/lib/balancer/gauges";
 import { pools } from "#/lib/gql/server";
 
 import { PoolStatsData, PoolTokens } from "../api/route";
-import { getBALPriceByRound, getTokenPriceByRound } from "./getBALPriceByRound";
+import { getBALPriceByRound, getTokenPriceByDate } from "./getBALPriceByRound";
 import { getPoolRelativeWeight } from "./getRelativeWeight";
 import { Round } from "./rounds";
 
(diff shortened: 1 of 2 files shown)


apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -95,8 +95,8 @@ async function calculateTokensStats(
   }, 0);
 
   const tokenPromises = poolTokenData.map(async (token, idx) => {
-    const tokenPrice = await getTokenPriceByRound(
-      Round.getRoundByNumber(roundId),
+    const tokenPrice = await getTokenPriceByDate(
+      Round.getRoundByNumber(roundId).endDate,
```
ribeirojose, line 99: 🤔 not sure I got why change this -- if anything we should change the API to get a start/end date instead of a fixed date, no?

apps/balancer-tools/src/app/apr/(utils)/getBALPriceByRound.ts
```diff
@@ -4,21 +4,43 @@ import { DefiLlamaAPI } from "#/lib/coingecko";
...
+        length:
+          Math.floor(
+            (round.endDate.getTime() - round.startDate.getTime()) /
+              MILLISECONDS_IN_DAY,
+          ) + 1,
+      },
+      (_, i) =>
+        getTokenPriceByDate(
+          new Date(round.startDate.getTime() + i * MILLISECONDS_IN_DAY),
+          "0xba100000625a3754423978a60c9317c58a424e3d",
```
ribeirojose, line 22: should be a constant -- this is the vault contract

ribeirojose (commented):
FYI -- defillama's API also accepts a searchWidth to get prices (https://defillama.com/docs/api)

ribeirojose (approved):

Rawallon:
> FYI -- defillama's API also accepts a searchWidth to get prices (https://defillama.com/docs/api)

I saw that, but does that mean that they will return avg value for a given width? I didn't see anything regarding that in their docs

PR #270, opened 2023-09-13
[APR] Validate getRate to calculate apr

luizakp:
This PR used hardcoded rETH rate provider address to validate if `getRate()` strategy would work

notes: This we be refactored on [this issue](https://linear.app/bleu-llc/issue/BAL-789/generic-function-to-getrate-for-different-tokens). The goal here is to unblock the integration with API


apps/balancer-tools/src/abis/rethRateProvider.ts
luizakp, line 1: @mendesfabio will it be possible to generalize this to just identify getRate() as a function?

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -0,0 +1,49 @@
...
+});
+
+const SECONDS_IN_DAY = 86400;
+
+export async function getTokenYieldByPoolId() {
+  const { currentRate, rate24hAgo } = await getDailyRateDiff();
+
+  const calc = (currentRate - rate24hAgo) / rate24hAgo;
+
+  const apr = ((1 + calc) ** 365 - 1) * 100;
```
luizakp, line 20: redo this to something like 
```suggestion
  const apr = ((currentRate - rate24hAgo) / e18)* 365
```

apps/balancer-tools/src/lib/gql.ts
```diff
@@ -61,3 +61,9 @@ export const internalBalances = {
   gql: (chainId: string) =>
     balancerSdks[networkFor(chainId)](internalBalances.client(chainId)),
 };
+
+export const blocks = {
+  client: clientFor(Subgraph.ChainsBlocks),
+  gql: (chainId: string) =>
+    balancerSdks[networkFor(chainId)](blocks.client(chainId)),
+};
```
luizakp, line 69: @ribeirojose I'm not sure about the blocks strategy. I touched on this before but I don't remember if we don't want to use subgraphs or not. Decided to use it here to make this validation easier

apps/balancer-tools/src/abis/rethRateProvider.ts
ribeirojose: Yup. Just remove the rest of the ABI and only leave the getRate definition here

apps/balancer-tools/src/lib/gql.ts
mendesfabio: some networks have consistent block times e.g. Ethereum 13s, Gnosis 5s, etc, so for those you may not need a Subgraph, take a look at the [SDK](https://github.com/search?q=repo:balancer/balancer-sdk%20blockTime&type=code) for a reference

luizakp: hmmmm got it but the way we build the gql service it will make the files for all the chains no matter what. would we prefer to have something like this to not use  unnecessary computational power?

luizakp: we can take a look this later on

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -0,0 +1,61 @@
...
+];
+
+export const publicClient = createPublicClient({
+  chain: mainnet,
+  transport: http(),
+});
+
+const SECONDS_IN_DAY = 86400;
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
luizakp, line 24: remove this on https://linear.app/bleu-llc/issue/BAL-791/interface-integrate-yield-apr-function-with-the-app

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -155,6 +161,8 @@ export async function calculatePoolStats({
           round.endDate.getTime() / 1000,
         ),
     ),
+    //TODO: on #BAL-795 use another strategy for cache using the poolId
+    getDataFromCacheOrCompute("yield APR", () => getTokenAprByPoolId(poolId)),
```
luizakp, line 165: Note:  currently, this is returning the same data for every round

PR #271, opened 2023-09-13
[APR] Add TokenAPR to historical chart

Rawallon:
image

# With mocked data

image

image


Rawallon:
Implements on top of #270

PR #285, opened 2023-09-18
[APR] getRateAtBlock for other chains

luizakp:
This PR adds getRateAtBlock for other chains.
This affects especially pools with `stMatic`.  `MaticX` case will be handled on BAL-781


apps/balancer-tools/src/app/apr/(utils)/chainsPublicClients.ts
```diff
@@ -0,0 +1,45 @@
...
+  | typeof base;
+
+export type ChainName =
+  | "ethereum"
+  | "polygon"
+  | "polygon-zkevm"
+  | "arbitrum"
+  | "avalanche"
+  | "gnosis"
+  | "base";
```
ribeirojose, line 28: couldn't we use the enums we already have? :/

apps/balancer-tools/src/app/apr/(utils)/poolsRateProvider.ts
```diff
@@ -0,0 +1,30 @@
...
+  },
+  {
+    poolAddress:
+      "0x577f6076e558818a5df21ce4acde9a9623ec0b4c000200000000000000000a64",
+    address: "0xeE652bbF72689AA59F0B8F981c9c90e2A8Af8d8f",
+    token: {
+      address: "0xfa68FB4628DFF1028CFEc22b4162FCcd0d45efb6",
+      symbol: "MaticX",
+    },
+    //TODO on BAL-781, MaticX does not update the rate daily.
```
ribeirojose, line 19: 🤔 interesting

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -81,37 +83,53 @@ function getAPRFromRate(
...
   return APR < 0 ? 0 : APR;
 }
 
 async function getPoolTokensRateProviders(chain: string, poolId: Address) {
   const data = await pools.gql(String(chain)).PoolRateProviders({ poolId });
 
-  if (!data.pool?.priceRateProviders?.length)
+  if (!data.pool?.priceRateProviders?.length) {
+    const poolRateProvider = manualPoolsRateProvider.find(
+      ({ poolAddress }) => poolAddress === poolId,
```
ribeirojose, line 94:
```suggestion
      ({ poolAddress }) => poolAddress.toLowerCase() === poolId.toLowerCase(),
```

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -81,37 +83,53 @@ function getAPRFromRate(
...
   }
-
   return APR < 0 ? 0 : APR;
 }
 
 async function getPoolTokensRateProviders(chain: string, poolId: Address) {
   const data = await pools.gql(String(chain)).PoolRateProviders({ poolId });
 
-  if (!data.pool?.priceRateProviders?.length)
+  if (!data.pool?.priceRateProviders?.length) {
```
ribeirojose, line 92: would we eventually use the manual approach to override bogus return values as well ? If so, we'd probably need to remove this if statement, right?

ribeirojose (dismissed):
👍 lgtm, but would try to avoid adding yet another pattern for chain naming/mapping 😅

apps/balancer-tools/src/app/apr/(utils)/chainsPublicClients.ts
luizakp: I tried but I got so complicated with the types that I decided to use this to make it work

luizakp:
This is only working for a single pool like: https://tools-monorepo-ro0533nq7-bleu-fi.vercel.app/apr/pool/polygon/0xeab6455f8a99390b941a33bbdaf615abdf93455e000200000000000000000a66

PR #320, opened 2023-09-22
[APR] refactor calculatepoolstats from round to date

Rawallon:
image


Rawallon:
To investigate: Sometimes the subgraph doesn't return a poolSnapshop

ribeirojose:
> To investigate: Sometimes the subgraph doesn't return a poolSnapshop

can we create a task for this? Fábio perhaps will be able to help with an answer

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -41,7 +43,7 @@ const fetchPoolAveragesInRange = withCache(
   ): Promise<[number, number, string, { symbol: string; balance: string }[]]> {
     const res = await pools.gql(network).poolSnapshotInRange({
       poolId,
-      from: from - SECONDS_IN_DAY,
+      from: from,
```
ribeirojose, line 46:
```suggestion
      from,
```

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -125,57 +126,58 @@ async function calculateTokensStats(
...
     ),
   ]);
 
   const tokens = await calculateTokensStats(
-    round,
+    endAt,
     pool.tokens,
     network,
     tokenBalance,
   );
```
ribeirojose, line 176: 🤔 wouldn't we also need startAt here?

apps/balancer-tools/src/app/apr/pool/(components)/HistoricalCharts/HistoricalChart.tsx
```diff
@@ -64,33 +59,6 @@ export default function HistoricalChartWrapper({
     y: [],
   });
 
-  const selectedRoundShape =
```
ribeirojose, line 67: 👍 so now we're only showing the data from the selected period. Makes sense to me!

apps/balancer-tools/src/app/apr/api/(utils)/fetchDataForPoolIdDateRange.ts
```diff
@@ -23,8 +22,11 @@ export async function fetchDataForPoolIdDateRange(
 
     while (attempts < MAX_RETRIES) {
       try {
-        const currentRound = Round.getRoundByDate(dayDate);
-        const data = await calculatePoolStats({ round: currentRound, poolId });
+        const data = await calculatePoolStats({
+          startAt: new Date(dayDate.getTime() - 24 * 60 * 60 * 1000),
+          endAt: dayDate,
```
ribeirojose, line 27: 🤔 why always calculate starting at a day before? Also -- shouldn't we change this signature to only get the unix timestamp altogether? 😅 it feels we're transforming this back and forth all the time

apps/balancer-tools/src/app/apr/(utils)/getBALPriceForDateRange.ts
```diff
@@ -20,27 +20,27 @@ const calculateDaysBetween = (startDate: Date, endDate: Date) =>
...
+        new Date(startAt.getTime() + i * MILLISECONDS_IN_DAY),
+        BAL_TOKEN_ADDRESS,
+        BAL_TOKEN_NETWORK,
+      ),
+    );
+    try {
+      const prices = await Promise.all(pricePromises);
+      return calculateAverage(prices);
+    } catch (error) {
+      // TODO: BAL-782 - Add sentry here
```
ribeirojose, line 37: 👌

apps/balancer-tools/src/app/apr/(utils)/getBALPriceForDateRange.ts
```diff
@@ -20,27 +20,27 @@ const calculateDaysBetween = (startDate: Date, endDate: Date) =>
...
-    console.error(`Error fetching BAL price between ${startAt} and ${endAt}`);
-    throw error;
-  }
-});
+export const getBALPriceForDateRange = withCache(
+  async function getBALPriceByRoundFn(startAt: Date, endAt: Date) {
+    const days = calculateDaysBetween(startAt, endAt);
+    const pricePromises = Array.from({ length: days }, (_, i) =>
+      getTokenPriceByDate(
+        new Date(startAt.getTime() + i * MILLISECONDS_IN_DAY),
```
ribeirojose, line 28: 😅  again, we should probably just make this ingest a timestamp

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -206,15 +207,54 @@ export async function calculatePoolStats({
...
-  round: Round,
+function getWeeksApart(startAt: Date, endAt: Date) {
+  const millisecondsInAWeek = 7 * 24 * 60 * 60 * 1000;
+  const differenceInMilliseconds = Math.abs(
+    startAt.getTime() - endAt.getTime(),
+  );
+  return Math.floor(differenceInMilliseconds / millisecondsInAWeek);
+}
+
+function calculateAPRForDateRange(
```
ribeirojose, line 218: this function is a bit messed up right? It's getting APR inputs from elsewhere and calculating the veBAL APR within it, we should refactor this

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
ribeirojose: but this way we're only outputting the same token stats for every day, no?

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -206,15 +207,54 @@ export async function calculatePoolStats({
...
+        return {
+          totalBalanceEmissions:
+            acc.totalBalanceEmissions + weeklyBalanceEmissions,
+          weekCount: acc.weekCount + 1,
+        };
+      },
+      { totalBalanceEmissions: 0, weekCount: 0 },
+    );
+
+    emissions = totalBalanceEmissions / weekCount;
```
ribeirojose, line 253: the emissions might not be the same for every week, they'll change every change of year. So if the period is between the yearly cliff this will return an average, instead of a cliff on the graph, which is not what should appear there. makes sense? we can split another task to rework this though

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
Rawallon: No since we fetch it like this for bal:
`https://coins.llama.fi/prices/historical/<timestamp>/ethereum:0xba100000625a3754423978a60c9317c58a424e3d?searchWidth=6h`

[22nd](https://coins.llama.fi/prices/historical/1695340800/ethereum:0xba100000625a3754423978a60c9317c58a424e3d?searchWidth=6h) - 3.14
[21rst](https://coins.llama.fi/prices/historical/1695254400/ethereum:0xba100000625a3754423978a60c9317c58a424e3d?searchWidth=6h) - 3.23
[20th](https://coins.llama.fi/prices/historical/1695168000/ethereum:0xba100000625a3754423978a60c9317c58a424e3d?searchWidth=6h) - 3.28

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
luizakp: https://linear.app/bleu-llc/issue/BAL-876/fix-totalbalanceemissions-calculations-when-fetching-on-year-change

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
luizakp: https://linear.app/bleu-llc/issue/BAL-877/refactor-apr-to-calculate-other-apr-inside-of-it

PR #328, opened 2023-09-25
[APR] Fix validation and filtering for Type and Network

Rawallon:


Rawallon:
Merging it since its a simple change.

PR #337, opened 2023-09-26
Cache only external queries instead of all intermediary functions

ribeirojose:


apps/balancer-tools/src/lib/cache.ts
```diff
@@ -164,7 +166,9 @@ export const withCache = <T, Args extends Array<unknown>>(
   return async (...args: Args) => {
     const serializedArgs = serializeArgs(args);
     const hashedFn = hashFunction(fn);
-    const cacheKey = `fn:${hashedFn}:${serializedArgs}`;
+    const cacheKey = `fn:${
+      fn.name ? `${fn.name}_` : ""
+    }${hashedFn}:${serializedArgs}`;
```
ribeirojose, line 171: this makes it easier locally to spot a cacheKey

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -38,71 +37,68 @@ const WEEKS_IN_YEAR = 52;
...
+    from,
+    to,
+  });
+
+  res.poolSnapshots = res.poolSnapshots.sort(
+    (a, b) => a.timestamp - b.timestamp,
+  );
+
+  if (res.poolSnapshots.length === 0) {
+    // Following Fabio's recomendation on #BAL-872
```
ribeirojose, line 57: This came from a rebase, I could have removed/added something that I shouldn't have here @Rawallon @luizakp

luizakp: tested here, everything working fine

PR #346, opened 2023-09-27
[APR] Handle when selected date is lower than pool created date

luizakp:
https://www.loom.com/share/94a8ae2d62524319a2c9125e3a618caa?sid=46906aee-ffa9-46f9-ac2f-ea8e79f5a9c4


Rawallon (approved):
Looks good, the redirect made me wonder if we shouldn't be doing that by the messaged received on the`fetcher`, since if you enter a URL with a bad pool name still shows Next Error (caused by throw on fetcher) eg http://localhost:3000/apr/pool/ethereum/bad%20name?startAt=09-27-2023&endAt=09-20-2023

PR #357, opened 2023-09-29
APR investigate why some pools get connection refused startat

Rawallon:
Before

https://github.com/bleu-fi/balancer-tools/assets/68453900/46770be7-5cfa-4909-8337-b2c3d1b7c6e5

After

https://github.com/bleu-fi/balancer-tools/assets/68453900/8581416a-ccbf-44a2-bc5b-f1d8e769c819

### DIFF
--- a/apps/balancer-tools/package.json
+++ b/apps/balancer-tools/package.json
@@ -51,6 +51,7 @@
     "lodash": "^4.17.21",
     "lodash.merge": "^4.6.2",
     "next": "^13.5.3",
+    "p-limiter": "^1.0.0",
     "plotly.js": "^2.26.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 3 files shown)


ribeirojose:
just leaving my two cents here: I suspect this is probably more of an issue in dev than in prod because in prod when we call multiple API endpoints that's spanning multiple lamba functions that each have their own IP and thus won't be rate limited (which I think is your assumption here). But would confirm that this phenomenon also occurs in prod taking a look at the logs.

ribeirojose (changes requested):
I'm blocking this because I don't think this is a solution for production. If we limit the number of pool x date combinations we query each time this request will timeout before all other requests are made, and we're thus always going to send an error to the user.

If the problem we're trying to solve here is development experience locally, I'd fallback to e.g. selecting the first n(say, 10) pools to ensure a smooth dev experience, and then think of what's being a blocker in production separately. FWIW I think we might've got to a point in which we need to think of alternative solutions here if this isn't scaling well. 😕

PR #367, opened 2023-10-03
[APR] Unify URLs

Rawallon:
I've noticed that in the past, we were redirecting users to "/apr?startAt=x&endAt=y" endpoint without including the default filter parameters. This unintentionally causes a "double redirect". Here's how it played out: Users would first get sent to "/apr?startAt=x&endAt=y," and then they'd get redirected again, this time with default values for parameters like minTvl, sort, and order. This extra redirection created unnecessary fetch requests.

On the flip side, when users didn't fill in the default values, they'd end up on the homepage without any sorting or ordering applied to the pool list table. This wasn't great for their overall user experience.

The goal of this pull request is twofold. First, it's meant to fix the problem I described. Second, it's designed to streamline the redirect process by handling it in a single function, so we don't have to do it repeatedly in different components.


luizakp:
not quite sure what this PR does hehehhehe @Rawallon can you provide more context on the description?

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
```diff
@@ -67,9 +70,36 @@ export default function getFilteredDateApiUrl(
...
+  const queryParams = generateQueryParams(startAt, endAt, searchParams);
+  return `${BASE_URL}/apr/api?${queryParams}${
     poolId ? `&poolId=${poolId}` : ""
-  }&${params}`;
+  }`;
+}
+
+export function generateRedirectUrlWithParams(
+  startAt: Date,
+  endAt: Date,
```
ribeirojose, line 92: shouldn't we just default to using a MMDDYYYY string instead of everywhere instead of passing it to date then passing it again back to this format? I notice we've done it a bunch of times across the code. Part of me also thinks using a timestamp would've made the whole implementation easier (although, admittedly, the links less readable)

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
```diff
@@ -67,9 +70,36 @@ export default function getFilteredDateApiUrl(
...
+  poolId?: string,
+) {
+  const queryParams = generateQueryParams(startAt, endAt, searchParams);
+  return `${BASE_URL}/apr/api?${queryParams}${
     poolId ? `&poolId=${poolId}` : ""
-  }&${params}`;
+  }`;
+}
+
+export function generateRedirectUrlWithParams(
```
ribeirojose, line 90:
```suggestion
export function generatePoolPageLink(
```

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
```diff
@@ -67,9 +70,36 @@ export default function getFilteredDateApiUrl(
...
+  startAt: Date,
+  endAt: Date,
+  searchParams?: SearchParams | null,
+  poolId?: string,
+) {
+  const queryParams = generateQueryParams(startAt, endAt, searchParams);
+  if (poolId) {
+    const network = networkFor(new Pool(poolId).network);
```
ribeirojose, line 98: if this pool doesn't have a gauge it won't have a network here, right?

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
```diff
@@ -47,14 +51,13 @@ function getFilterDataFromParams(searchParams: SearchParams) {
   return result;
 }
 
-export default function getFilteredDateApiUrl(
+function generateQueryParams(
```
ribeirojose, line 54: there's some duplication in this file that made things a bit harder to parse, sth to keep an eye on

ribeirojose (approved):

apps/balancer-tools/src/app/apr/(utils)/getFilteredApiUrl.ts
Rawallon: All pools have network, regardless of gauge, We add it on the pre-build.js [link to file](https://github.com/bleu-fi/balancer-tools/blob/main/apps/balancer-tools/src/data/pools-without-gauge.json#L3).

PR #375, opened 2023-10-04
[APR] Robust token yield apr

Rawallon:
On prod some pools are returning "0" on token yield where it shouldn't be [[example](https://balancer-tools.bleu.fi/apr/pool/ethereum/0x1ee442b5326009bb18f2f472d3e0061513d1a0ff000200000000000000000464?startAt=09-30-2023&endAt=10-03-2023)]. I couldn't replicate on dev., so I went for trying to make it more robust, and warn us whenever an error occurs.


luizakp:
@Rawallon this doesn't need to be merged for you to debug it, right?

PR #377, opened 2023-10-04
[APR] Refactor trimTrailingValues to remove all valueToTrim values

Rawallon:
Before it would return an array, if no other value was found

image


luizakp (approved):
from what I have checked this doesn't change the functionalities on pool simulator

PR #382, opened 2023-10-05
[APR] Fix token yield pt.1

Rawallon:
On prod some pools are returning "0" on token yield where it shouldn't be [[example](https://balancer-tools.bleu.fi/apr/pool/ethereum/0x1ee442b5326009bb18f2f472d3e0061513d1a0ff000200000000000000000464?startAt=09-30-2023&endAt=10-03-2023)]. I couldn't replicate on dev., so I went for trying to make it more robust, and warn us whenever an error occurs.


apps/balancer-tools/src/app/apr/(utils)/chainsPublicClients.ts
```diff
@@ -5,6 +5,7 @@ import {
   base,
   gnosis,
   mainnet,
+  optimism,
```
ribeirojose, line 8: 🙏

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -73,14 +98,34 @@ async function getAPRFromRateProviderInterval(
       timeEnd,
       chainName,
     );
+    Sentry.addBreadcrumb({
+      category: "auth",
+      message: "endRate: " + endRate,
+      level: "info",
+    });
+    Sentry.addBreadcrumb({
+      category: "auth",
```
ribeirojose, line 107: why auth?

apps/balancer-tools/src/app/apr/(utils)/tokenYield.ts
```diff
@@ -56,13 +77,17 @@ async function getAPRFromRateProviderInterval(
...
   chainName: ChainName,
+  poolId: string,
 ) {
   if (
     timeEnd >= 1692662400 && // pool vunerability was found on August 22
     vunerabilityAffecteRateProviders.some(
       ({ address }) => address === rateProviderAddress,
     )
   ) {
+    Sentry.captureMessage(
```
ribeirojose, line 88: do we need to add a new entry to sentry here because of this?

apps/balancer-tools/src/lib/cache.ts
```diff
@@ -29,6 +29,17 @@ const hashFunction = <
...
+  if (hashCache[input]) return hashCache[input];
+
+  const hash = crypto.createHash("sha256");
+  hash.update(input);
+  const hashValue = hash.digest("hex");
+
+  hashCache[input] = hashValue;
+  return hashValue;
+};
+
```
luizakp, line 42: @ribeirojose do you think this will be an issue? I was trying to solve the case where we have to retry the snapshot because I get something like:

```
Cache error during write: Error: ENAMETOOLONG: name too long, open './.cache/cache_localhost_fn_bound poolSnapshotInRange_0b57ab05538ff09a83dc6159fb7e5e46c4188ee02fe567ce5e1a1ef6823d6e30_{_poolId___0xedf085f65b4f6c155e13155502ef925c9a756003000200000000000000000123_,_timestamp__[1696118400,1696204800,1696291200,1696377600,1696464000,1696550400,1696636800,1696723200,1696809600]}.json





[11:04](https://bleu-llc.slack.com/archives/D04EV3EK524/p1697033086510059)
```

luizakp: it tries to save all the attempted timestamps on timestamp

ribeirojose: I don't think it'd be an issue, but this might make it nearly impossible to decode keys in Redis... I don't see that as being a major issue though since I don't think it's been in our primary flow to accomplish that

luizakp: but it kind of is a major issue, no? All the days of all possible pools that we have a retry are not saved on cache

ribeirojose: can you articulate your concern more? I'm not sure I follow. What I meant is that hashing the keys makes it harder  for us as devs to go into redis and inspect an individual value because we'd have to start with the unhashed key and apply the algorithm to it do find the key that we'd then be able to find in Redis. It's just a matter of how hard it'd become to query redis

luizakp: aaaa sorry I misunderstood what you were saying hehehehhe I understood that not having the cache for those days wasn't a major issue. Now I got it that hashing the args isn't a major issue

Rawallon: In that case couldn't we change `serializeArgs` fn so that when receiving an array only saves the first and last member, that way he return would become `{_poolId___0xedf085f65b4f6c155e13155502ef925c9a756003000200000000000000000123_,
_timestamp__[1696118400...1696809600]}`

PR #402, opened 2023-10-10
[APR] Remove token price on calculate pool stats and delete pool token table

luizakp:
This PR:
- deletes pool token table
- remove the token price on calculate pools stats (token stats aren't necessary anymore because we don't use pool token table)


ribeirojose (approved):

ribeirojose:
👍 good initiative!

PR #418, opened 2023-10-16
[Balancer APR] add filter params to base subgraph query

ribeirojose:
This PR:
- remove blocks subgraph
- use coin.llama block API
- handle proxy contract as rate provider
- refactor to get `existingPoolForDate` on `fetchForDateRange`


apps/balancer-tools/pre-build.js
```diff
@@ -70,7 +70,7 @@ const EXISTING_NETWORKS = {
 const POOLS_WITHOUT_GAUGE_QUERY = `
 query PoolsWherePoolType($skip: Int!) {
   pools(
-    where: { totalLiquidity_gt: 5000 }
+    where: { totalLiquidity_gt: 10000 }
```
ribeirojose, line 73: this should probably be removed altogether

luizakp: We have some validations on `gauge.ts` and `validate.ts` that use `POOLS_WITHOUT_GAUGES`. I will think about how we can remove this. My only fear is that it will try to perform the validations before `APRPools` query is done

ribeirojose: gotcha. fwiw this could have virtually all pools (so no totalLiquidity filter) if it's only being used for getting the pool created date

apps/balancer-tools/src/app/apr/api/(utils)/validate.ts
```diff
@@ -96,8 +86,18 @@ export const QueryParamsSchema = z
...
+      {
+        message: "minTvl should be at least 10,000",
+      },
+    ),
+    maxTvl: OptionalNullableFloat.refine(
+      (value) => value === null || (value !== null && value <= 10000000000),
+      {
+        message: "maxTvl should not be greater than 10,000,000,000",
+      },
+    ),
```
ribeirojose, line 100: why add these validations here? They'll only get in the way IMO 😅

apps/balancer-tools/src/utils/fetcher.ts
```diff
@@ -5,9 +5,12 @@ export async function fetcher<JSON = unknown>(
   const response = await fetch(input, init);
 
   if (!response.ok) {
-    // eslint-disable-next-line no-console
-    console.error(await response.text());
-    throw new Error("Network response was not ok");
+    const errorBody = await response.json();
```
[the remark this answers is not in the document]
ribeirojose: could also check for in response.json() for `"error"` or `"errors"` key, since GQL queries that break won't return an non 2xx status but will come with those keys in the response

apps/balancer-tools/src/app/apr/api/(utils)/filter.ts
```diff
@@ -59,8 +59,8 @@ const conditions: ConditionTypes = {
     value!
       .map((pType) => pType.toLowerCase())
       .includes(pool.type.toLowerCase()),
-  minTvl: (pool, value) => pool.tvl >= value!,
-  maxTvl: (pool, value) => pool.tvl <= value!,
+  minTvl: (pool, value) => pool.tvl >= value! && value! >= 10_000,
+  maxTvl: (pool, value) => pool.tvl <= value! && value! <= 10_000_000_000,
```
ribeirojose, line 63: if you want to keep this here (which I don't suggest, you're coupling this to a context that it shouldn't need to know about), would centralize these constants somewhere and replace it for a reference to them so that, if you ever want to change this, you don't need to go about doing it everywhere

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -229,14 +226,62 @@ const getRateAtBlock = withCache(async function getRateAtBlockFn(
...
     );
-    Sentry.captureException(e);
-    throw e;
+    if (actualRateProvider) {
+      return getRateAtBlock(chainName, actualRateProvider, poolId, blockNumber);
+    } else {
+      // eslint-disable-next-line no-console
+      console.error(
+        `Error fetching rate for ${rateProviderAddress} chain ${chainName} poolId ${poolId}`,
+      );
```
ribeirojose, line 247:
```suggestion
      // eslint-disable-next-line no-console
      console.error(
        `Error fetching rate for ${rateProviderAddress} chain ${chainName} poolId ${poolId}. This is probably a Linear pool and the call reverted, in which case we can assume value 0.`,
      );
      return 0
```

apps/balancer-tools/src/app/apr/api/(utils)/fetchForDateRange.ts
```diff
@@ -1,21 +1,105 @@
...
+
+const fetchPoolsFromNetwork = async (
+  network: string,
+  params: ReturnType<typeof QueryParamsSchema.safeParse>,
+  skip = 0,
+) => {
+  if (!params.success) return [];
+
+  const maxTvl = params.data.maxTvl || 10_000_000_000;
+  const minTvl = params.data.minTvl || 10_000;
```
luizakp, line 23: sorry didn't pay attention this was an `or`

apps/balancer-tools/src/app/apr/api/(utils)/validate.ts
luizakp: https://github.com/bleu-fi/balancer-tools/pull/418#discussion_r1360999805 hehehe sorry

luizakp (approved):
👍

PR #420, opened 2023-10-16
[APR] Investigating Prod bugs pt.1

luizakp:
This PR:
- add more bb pools
- increase the maxDuration of the server function
- Add missing suspense on header

### DIFF
--- a/apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
@@ -166,10 +166,6 @@ async function getFeeAprForDateRange(
   });
 
   if (res.poolSnapshots.length === 0) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      `No data found for pool ${poolId}(${network}) in range ${from} - ${to}`,
-    );
     return [0, 0];
   }
 
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -58,6 +58,9 @@ export async function fetchPoolAveragesForDateRange(
     chosenData = retryGQL.poolSnapshots
       .sort((a, b) => a.timestamp - b.timestamp)
       .slice(-1)[0];
+
+    // if the retry was needed then the volume on that day is 0
+    chosenData.swapVolume = 0;
   }
 
   const liquidity = Number(chosenData.liquidity);
--- a/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
@@ -46,21 +46,12 @@ export async function fetchPoolAveragesForDateRange(
   let chosenData = res.poolSnapshots[0];
 
   if (res.poolSnapshots.length == 0) {
-    console.warn(
-      "No return on poolSnapshots, trying to fetch in a few days before",
-    );
     const retryGQL = await pools.gql(network).poolSnapshotInRange({
       poolId,
       timestamp: generateDateRange(from - SECONDS_IN_DAY * 7, to),
     });
 
     if (retryGQL.poolSnapshots.length === 0) {
-      console.error(
-        `No return on retrying poolSnapshots. ${poolId}(${network}) in range ${new Date(
-          from * 1000,
-        )} - ${new Date(to * 1000)}`,
-      );
-
       // TODO: Throw error here and handle outside of it.
       return [0, 0, "", 0];
     }
--- a/apps/balancer-tools/src/app/apr/(utils)/vunerabilityAffectedPool.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/vunerabilityAffectedPool.ts
@@ -64,4 +64,12 @@ export const vunerabilityAffecteRateProviders: {
     address: "0x8025586ac5fb265a23b9492e7414beccc2059ec3",
     token: "bb-rfWBTC",
   },
+  {
+    address: "0x60D604890feaa0b5460B28A424407c24fe89374a",
+    token: "bb-a-WETH",
+  },
+  {
+    address: "0x7275c131b1f67e8b53b4691f92b0e35a4c1c6e22",
+    token: "bb-a-WAVAX",
+  },
 ];
--- a/apps/balancer-tools/src/app/apr/(utils)/vunerabilityAffectedPool.ts
+++ b/apps/balancer-tools/src/app/apr/(utils)/vunerabilityAffectedPool.ts
@@ -48,4 +48,20 @@ export const vunerabilityAffecteRateProviders: {
     token: "bb-e-USDT",
     address: "0x3c640f0d3036ad85afa2d5a9e32be651657b874f",
   },
+  {
+    token: "bb-a-WMATIC",
+    address: "0xb0c830dceb4ef55a60192472c20c8bf19df03488",
+  },
+  {
+    token: "bb-a-WETH",
+    address: "0xad28940024117b442a9efb6d0f25c8b59e1c950b",
+  },
+  {
+    address: "0x48ace81c09382bfc08ed102e7eadd37e3b049752",
+    token: "bb-rfWSTETH",
+  },
+  {
+    address: "0x8025586ac5fb265a23b9492e7414beccc2059ec3",
+    token: "bb-rfWBTC",
+  },
 ];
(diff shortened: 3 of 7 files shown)


apps/balancer-tools/src/app/apr/layout.tsx
```diff
@@ -15,12 +16,14 @@ export const metadata: Metadata = {
...
-        endButton={<HeaderEndButton />}
-      />
+      <React.Suspense fallback={<Spinner />}>
+        <Header
+          linkUrl={"/apr"}
+          title={APP_DISPLAY_NAME}
+          imageSrc={"/assets/balancer-symbol.svg"}
+          endButton={<HeaderEndButton />}
+        />
+      </React.Suspense>
```
ribeirojose, line 26: 😮

apps/balancer-tools/src/utils/fetcher.ts
```diff
@@ -1,3 +1,5 @@
+export const maxDuration = 300;
```
ribeirojose, line 1: this is not a route, so this shouldn't do anything

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
```diff
@@ -166,10 +166,6 @@ async function getFeeAprForDateRange(
   });
 
   if (res.poolSnapshots.length === 0) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      `No data found for pool ${poolId}(${network}) in range ${from} - ${to}`,
-    );
```
ribeirojose, line 172: were we still hitting this many times?

ribeirojose: This just made me realize this is a similar error case from the below, right? Why don't we treat them the same? I think this might lead to inconsistent results. Did we get to the root cause of why some snapshots are missing? Finding that out is why these logs have been posted here.

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -46,21 +46,12 @@ export async function fetchPoolAveragesForDateRange(
...
       timestamp: generateDateRange(from - SECONDS_IN_DAY * 7, to),
     });
 
     if (retryGQL.poolSnapshots.length === 0) {
-      console.error(
-        `No return on retrying poolSnapshots. ${poolId}(${network}) in range ${new Date(
-          from * 1000,
-        )} - ${new Date(to * 1000)}`,
-      );
-
```
ribeirojose, line 63: same comments from above.

apps/balancer-tools/src/app/apr/(utils)/vunerabilityAffectedPool.ts
```diff
@@ -48,4 +48,28 @@ export const vunerabilityAffecteRateProviders: {
...
+    token: "bb-rfWBTC",
+  },
+  {
+    address: "0x60D604890feaa0b5460B28A424407c24fe89374a",
+    token: "bb-a-WETH",
+  },
+  {
+    address: "0x7275c131b1f67e8b53b4691f92b0e35a4c1c6e22",
+    token: "bb-a-WAVAX",
+  },
```
ribeirojose, line 74: 👍

apps/balancer-tools/src/app/apr/layout.tsx
ribeirojose: just curious about how you found this out, and what part of this component was throwing an error/loading external data?

apps/balancer-tools/src/app/apr/api/route.ts
```diff
@@ -10,6 +10,9 @@ import { filterPoolStats } from "./(utils)/filter";
 import { limitPoolStats, Order, sortPoolStats } from "./(utils)/sort";
 import { QueryParamsSchema } from "./(utils)/validate";
 
+export const dynamic = "force-dynamic";
```
ribeirojose, line 13: Not sure adding `"force-dynamic"` here does much. Did you take a look into the docs/notice any difference? I'd have added it to the root layout, not the here.

ribeirojose (dismissed):
I wish we had a root cause for the missing snapshots, and a potential list of the ones we found out to be missing with reproducible results. Apart from that, approving to unblock you but think we need to address the comments.

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
luizakp: They are the same error, that's why I removed them from the console warn. We were warned this 3x which was spamming the console and making it harder to debug other things xD 

But the root cause of missing snapshots is not an error, is just how the snapshot works... If no swap was made, then there is no snapshot for that day. Here is an example I looked at yesterday: 

The query is empty, and that is true on balancer FE too

https://app.balancer.fi/#/ethereum/pool/0x6ae5a7857aad26722cb02cc168e38c52e0e4e45d0000000000000000000005dd
[Query on the subgraph](https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2/graphql?query=++query+poolSnapshotInRange($poolId:+String!,+$timestamp:+%5BInt!%5D)+%7B%0A++++poolSnapshots(%0A++++++where:+%7B+pool_in:+%5B$poolId%5D,+timestamp_in:+$timestamp+%7D%0A++++++orderBy:+timestamp%0A++++++orderDirection:+asc%0A++++)+%7B%0A++++++pool+%7B%0A++++++++id%0A++++++++address%0A++++++++name%0A++++++++poolType%0A++++++++symbol%0A++++++++tokens+%7B%0A++++++++++symbol%0A++++++++++balance%0A++++++++%7D%0A++++++%7D%0A++++++amounts%0A++++++totalShares%0A++++++swapVolume%0A++++++protocolFee%0A++++++swapFees%0A++++++liquidity%0A++++++swapsCount%0A++++++holdersCount%0A++++++timestamp%0A++++%7D%0A++%7D)

ribeirojose: so if that's the case, the volume should be 0 for that day, and the liquidity the same from the previous day, right?

luizakp:
> I wish we had a root cause for the missing snapshots, and a potential list of the ones we found out to be missing with reproducible results. Apart from that, approving to unblock you but think we need to address the comments.

I can make a list and search for 5-10 pools to make sure the case I presented here is the same for other pools so we are confident that those pools really don't have snapshots or something is missing on our end. WDYT? @ribeirojose

ribeirojose:
> > I wish we had a root cause for the missing snapshots, and a potential list of the ones we found out to be missing with reproducible results. Apart from that, approving to unblock you but think we need to address the comments.
> 
> I can make a list and search for 5-10 pools to make sure the case I presented here is the same for other pools so we are confident that those pools really don't have snapshots or something is missing on our end. WDYT? @ribeirojose

Initially I'd say yes, but I think [your comment here](https://github.com/bleu-fi/balancer-tools/pull/420#discussion_r1362258884) made me realize there's a bunch of cases where the subgraph won't have the pool snapshot at all, so we have to treat that on a one-off basis 😓  think this is something we should keep in mind for the future but Idt there's much use for that list right now. Sorry for not having the full picture in mind!

apps/balancer-tools/src/app/apr/api/route.ts
luizakp: I'm dropping this for now. The docs say that this will miss all the cache 🫠

ribeirojose: FYI just to make i clear this would miss all Vercel cache, not our own

apps/balancer-tools/src/app/apr/layout.tsx
luizakp: `HeaderEndButton` has `const searchParams = useSearchParams();` that's why the component needs a suspense

ribeirojose: so the suspense should be only within the HeaderEndButton, no? 👁️ 👄 👁️

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
luizakp: Yeah, that's why we do the [retry](https://github.com/bleu-fi/balancer-tools/blob/6dbf917b357275f899c8d5b3ba01a6cb7471e6f9/apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts#L51). We missed the fact that the volume should be 0 hehehhehe thanks! (Extra question: I was thinking yesterday If it was possible to get this from the last cached value instead of querying it again but not sure how this would behave when we have multiple days not cached)

But the code we are commenting on right now is the feeApr and collected Fee, if there is no snapshot they should be indeed 0

apps/balancer-tools/src/app/apr/layout.tsx
luizakp: heheheheh true

PR #430, opened 2023-10-18
[APR] Not cache current day on Balancer subgraph requests

luizakp:
This PR is a possible solution for the following problem: queries like poolSnapshot update the result throughout the day, so we don't want the cache the current day. 

why follow this approach? was how I solved the problem without interfering with the cache function itself. Tried to play around with the args on `wrapWithCache` but seemed like it would be just another entrance for bugs.


apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -38,17 +39,17 @@ export async function fetchPoolAveragesForDateRange(
...
 
   let chosenData = res.poolSnapshots[0];
 
   if (res.poolSnapshots.length == 0) {
     const retryGQL = await pools.gql(network).poolSnapshotInRange({
       poolId,
-      timestamp: generateDateRange(from - SECONDS_IN_DAY * 7, to),
+      timestamp: generateDateRange(
+        from - SECONDS_IN_DAY * 7,
+        to - SECONDS_IN_DAY,
```
luizakp, line 51: If the original `to` already failed then there is no need to try it again

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
```diff
@@ -0,0 +1,43 @@
...
+        now.getUTCFullYear(),
+        now.getUTCMonth(),
+        now.getUTCDate(),
+        0,
+        0,
+        0,
+        0,
+      ),
+    ),
+  );
```
luizakp, line 28: This is getting today but at 00:00 UTC, which the hour snapshot expects

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
```diff
@@ -0,0 +1,43 @@
...
+    res = await poolsWithoutCache.gql(network).poolSnapshotInRange({
+      poolId,
+      timestamp: generateDateRange(extendedFrom, to),
+    });
+  } else {
+    res = await pools.gql(network).poolSnapshotInRange({
+      poolId,
+      timestamp: generateDateRange(extendedFrom, to),
+    });
+  }
```
luizakp, line 40: not the biggest fan of these... but it gets the job done. Happy to hear other solutions

luizakp:
tvl seem wrong by one day

image
image

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
luizakp, line 1: tvl seem wrong by one day

image
image

luizakp: solved:
https://github.com/bleu-fi/balancer-tools/pull/430/commits/6143ed1b3f5ecd776814b54abf4123610d53fc70
image
image

luizakp:
skipping cache didn't work on prod 😢 https://www.loom.com/share/c307ad51fd4348ca9aedefefd48be851?sid=dea4a377-4fa8-45c1-8b3e-004ad8084ab0

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
```diff
@@ -2,17 +2,17 @@ import { Address } from "@bleu-balancer-tools/utils";
 import * as Sentry from "@sentry/nextjs";
 
 import * as balEmissions from "#/lib/balancer/emissions";
-import { pools } from "#/lib/gql/server";
+import { poolsWithoutCache } from "#/lib/gql/server";
```
ribeirojose, line 5: honestly I think it was my mistake having the default as cached. `#would` reverse this, so implement "poolsWithCache" and use it where needed, and leave the default not having cached.

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (to === currentDateUTC) {
+    res = await poolsWithoutCache.gql(network).poolSnapshotInRange({
+      poolId,
+      timestamp: generateDateRange(extendedFrom, to),
+    });
+  } else {
+    res = await pools.gql(network).poolSnapshotInRange({
+      poolId,
+      timestamp: generateDateRange(extendedFrom, to),
```
ribeirojose, line 38: I'm not sure if I said this, but I wonder if we're having errors because we're filtering using timestamps we generated in a list instead of more loosely using `timestamp_gte` and `timestamp_lte` filters. I thought that was what we were using but could be remembering this wrong. Do you know why we're not using that approach?

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
ribeirojose: you could do something like 
```
strategy = to === currentDateUTC ? poolsWithoutCache : pools;
const res = await strategy.gql(network).poolSnapshotInRange({
  poolId,
  timestamp: generateDateRange(extendedFrom, to),
});
```

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
```diff
@@ -0,0 +1,43 @@
...
+export async function fetchPoolSnapshots(
+  to: number,
+  extendedFrom: number,
+  network: string,
+  poolId: string,
+): Promise<PoolSnapshotInRangeQuery> {
+  let res: PoolSnapshotInRangeQuery;
+
+  const now = new Date();
+  const currentDateUTC = dateToEpoch(
```
ribeirojose, line 16:
```suggestion
  const utcMidnightTimestampOfCurrentDay = dateToEpoch(
```

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
```diff
@@ -0,0 +1,43 @@
...
+    new Date(
+      Date.UTC(
+        now.getUTCFullYear(),
+        now.getUTCMonth(),
+        now.getUTCDate(),
```
ribeirojose, line 21: 👌

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
ribeirojose, line 16: if we were using the solution described by https://github.com/bleu-fi/balancer-tools/pull/430/files#r1365501152 we wouldn't need to do this since all UNIX timestamps are in UTC time

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
```diff
@@ -177,7 +173,7 @@ async function getFeeAprForDateRange(
     sortedSnapshots.length <= 1 &&
     sortedSnapshots[sortedSnapshots.length - 1].timestamp <= to
   ) {
-    const currentData = await pools.gql(network).Pool({ poolId });
+    const currentData = await poolsWithoutCache.gql(network).Pool({ poolId });
```
ribeirojose, line 176: just making a note re. https://github.com/bleu-fi/balancer-tools/pull/430/files#r1365495857

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
ribeirojose: IMO there's no need to add this complication though

ribeirojose (commented):
considering this again, I think I might have something missing in my mental model. Aren't poolSnapshots generated every day at midnight UTC? Why is this even an issue then? I can see how using uncached results could be an issue when querying against the pool entity, which gets updated after every swap/join/exit. But poolSnapshots are (supposedly) never updated, so I'm not sure how this can change anything. Can you help me identify where I'm considering things wrong here?

packages/gql/src/balancer/pools.ts
```diff
@@ -84,7 +84,7 @@ export const poolSnapshotInRange = gql`
     poolSnapshots(
       where: { pool_in: [$poolId], timestamp_in: $timestamp }
       orderBy: timestamp
-      orderDirection: asc
```
ribeirojose, line 87: why change the direction here?

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
luizakp: IMO they both do the same, just different versions to achieve the same thing. This a change during the refactor you made https://github.com/bleu-fi/balancer-tools/blob/81745fb789394fe760ac2de431d50bd7c74e8c11/packages/gql/src/balancer/pools.ts and I thought it was well put as the timestamps are specific

apps/balancer-tools/src/app/apr/(utils)/fetchPoolSnapshots.ts
luizakp: not sure if I follow, even using https://github.com/bleu-fi/balancer-tools/pull/430/files#r1365501152 it would be needed to get `utcMidnightTimestampOfCurrentDay`

packages/gql/src/balancer/pools.ts
luizakp: Here: https://github.com/bleu-fi/balancer-tools/pull/430#discussion_r1364572679

when fetching `from` 17/10 `to` 18/10 it was returning day 17/10 first. Which resulted in catching 17/10 instead of 18/10 and using the dates wrongly by one day

luizakp:
> considering this again, I think I might have something missing in my mental model. Aren't poolSnapshots generated every day at midnight UTC? Why is this even an issue then? I can see how using uncached results could be an issue when querying against the pool entity, which gets updated after every swap/join/exit. But poolSnapshots are (supposedly) never updated, so I'm not sure how this can change anything. Can you help me identify where I'm considering things wrong here?

yup hahahhah I think you just forgot that poolSnapshots are updated after every swap/join/exit as well, so they are not generated every day at midnight. They are generated at the first swap/join/exit of the day and saved on that day day at midnight UTC just as a convention. Therefore, during the current day the poolSnapshots updates.

Now the issue was:  we were using cache and let's say you used the app today (19/10) at 8am. A certain pool has $20 as a collected fee during that time. The data from 8am was being saved as the data for the pools for all of today. So even at today 9pm if you entered the same pool, it would hit cache and return $20 as a collected fee, while if you query the pool on the subgraph at 9pm the collected fee is $45. The problem gets bigger as time passes because on the next day (20/10) the date 19/10 will also cache hit but for the data collected which is $20, it would be different from what was really collected that day.

let me know if that was confusing kkkkkk

PR #437, opened 2023-10-19
[APR] Add fathom to balancer toolsbleufi

luizakp:
This PR followed the new docs for next 13 https://github.com/derrickreimer/fathom-client#nextjs

### DIFF
--- a/apps/balancer-tools/src/components/Fathom.tsx
+++ b/apps/balancer-tools/src/components/Fathom.tsx
@@ -1,0 +1,39 @@
+"use client";
+
+import { load, trackPageview } from "fathom-client";
+import { usePathname, useSearchParams } from "next/navigation";
+import { Suspense, useEffect } from "react";
+
+const appUrl = "balancer-tools.bleu.fi";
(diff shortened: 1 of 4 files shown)


apps/balancer-tools/src/components/Fathom.tsx
```diff
@@ -0,0 +1,39 @@
+"use client";
+
+import { load, trackPageview } from "fathom-client";
+import { usePathname, useSearchParams } from "next/navigation";
+import { Suspense, useEffect } from "react";
+
+const appUrl = "balancer-tools.bleu.fi";
+const fathomId = "HLWGNXQK";
```
luizakp, line 8: @ribeirojose not sure it's okay to leave it like this. Thought about adding to an env, but both official docs just say to add the fathom id. So I decided to not over complicate at first, but if necessary change this

ribeirojose: this is going to be in the frontend anyways, so no worries

PR #439, opened 2023-10-20
[APR] APR Pools query with tokens

luizakp:
closes [BAL-1160](https://linear.app/bleu-llc/issue/BAL-1160/fix-token-filter)

### DIFF
--- a/packages/gql/src/balancer/pools.ts
+++ b/packages/gql/src/balancer/pools.ts
@@ -176,7 +176,7 @@ export const aprPools = gql`
     $skip: Int!
     $createdBefore: Int
     $limit: Int
-    $tokens: [Bytes!]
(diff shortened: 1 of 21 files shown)


packages/gql/src/balancer/pools.ts
```diff
@@ -176,17 +176,17 @@ export const aprPools = gql`
...
     $maxTvl: BigDecimal
     $block: Int
   ) {
     pools(
       where: {
         createTime_lte: $createdBefore
-        tokensList_contains: $tokens
         totalLiquidity_gte: $minTvl
         totalLiquidity_lte: $maxTvl
+        tokens_: { symbol_in: $tokens }
```
ribeirojose, line 189: 🤯

PR #446, opened 2023-10-25
[APR] Review yield apr

luizakp:
- FIx yield APR calc for weighted and composable pools
explaining:
https://www.loom.com/share/6caef1e19dca40c9af8915ec1214d856?sid=c163f68c-a0a9-443c-99fe-576aaba0da16


packages/gql/src/balancer/pools.ts
```diff
@@ -110,6 +110,39 @@ export const poolSnapshotInRange = gql`
...
+          priceRate
+          weight
+          isExemptFromYieldProtocolFee
+        }
+      }
+      timestamp
+    }
+  }
+`;
+
```
ribeirojose, line 145: why not simply add this to the existing poolSnapshotInRange query?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -76,14 +83,21 @@ export async function getPoolTokensAprForDateRange(
...
+}: {
+  rateProviderAddress: Address;
+  timeStart: number;
+  timeEnd: number;
+  chainName: ChainName;
+  poolId: string;
+  tokenAddress: Address;
+}) {
   if (
     timeEnd >= 1692662400 && // pool vunerability was found on August 22
```
ribeirojose, line 102: we should add this to a const

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -55,14 +62,14 @@ export async function getPoolTokensAprForDateRange(
             return {
               address: tokenAddress,
               symbol,
-              yield: await getAPRFromRateProviderInterval(
-                rateProviderAddress as Address,
-                startAt,
-                endAt,
+              yield: await getAPRFromRateProviderInterval({
```
ribeirojose, line 65: I really don't care about it all that much, but if we're going to change one signature to be an object we might as well change all of them to ensure the same pattern. Personally I think this makes the code more verbose but I'm ok with that as long as we're consistent/there's a reason for each exception. wdyt?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -55,14 +62,14 @@ export async function getPoolTokensAprForDateRange(
...
+                rateProviderAddress: rateProviderAddress as Address,
+                timeStart,
```
ribeirojose, line 67: ☝️ same goes for arg names

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -55,14 +62,14 @@ export async function getPoolTokensAprForDateRange(
             return {
               address: tokenAddress,
               symbol,
-              yield: await getAPRFromRateProviderInterval(
-                rateProviderAddress as Address,
-                startAt,
-                endAt,
+              yield: await getAPRFromRateProviderInterval({
+                rateProviderAddress: rateProviderAddress as Address,
```
ribeirojose, line 66: what do we gain by enforcing this here or instead of the function accepting a string instead?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+    poolId,
+    timestamp: generateDateRange(timeStart, timeEnd),
+  });
+
+  let pool: Pool;
+
+  if (rangeData.poolSnapshots.length > 0) {
+    pool = rangeData.poolSnapshots[0].pool;
+  } else {
+    const currentPoolData = await pools.gql(chain).Pool({ poolId });
```
ribeirojose, line 352: Just to keep in mind -- I think we would've already fetched this in that Promise.all where we get data from all APR functions

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+  let apr;
+
+  if (
+    pool.poolType === PoolTypeEnum.PHANTOM_STABLE ||
+    (pool.poolType === PoolTypeEnum.WEIGHTED &&
+      (pool.poolTypeVersion ?? 0) >= 2)
+  ) {
+    const isExemptFromYieldProtocolFee = pool?.tokens.find(
+      ({ address }) => address === tokenAddress,
+    )?.isExemptFromYieldProtocolFee;
```
ribeirojose, line 372: this assumes only one of the tokens would be exempt from YieldProtocolFee, but there could be multiple for the same pool, no? If not, I think there should be a comment here explaining why

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+
+  let apr;
+
+  if (
+    pool.poolType === PoolTypeEnum.PHANTOM_STABLE ||
+    (pool.poolType === PoolTypeEnum.WEIGHTED &&
+      (pool.poolTypeVersion ?? 0) >= 2)
+  ) {
+    const isExemptFromYieldProtocolFee = pool?.tokens.find(
+      ({ address }) => address === tokenAddress,
```
ribeirojose, line 371: ~this assumes only one of the tokens would be exempt from YieldProtocolFee, but there could be multiple for the same pool, no? If not, I think there should be a comment here explaining why~

nvm, read this quickly and saw we're not finding by `isExemptFromYieldProtocolFee` but just getting the token address that's been passed to the function

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+    )?.isExemptFromYieldProtocolFee;
+
+    if (isExemptFromYieldProtocolFee) {
+      apr = tokenYield;
+    } else {
+      //source: https://github.com/balancer/balancer-sdk/blob/f4879f06289c6f5f9766ead1835f4f4b096ed7dd/balancer-js/src/modules/pools/apr/apr.ts#L134
```
ribeirojose, line 377: 👍

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+    const currentPoolData = await pools.gql(chain).Pool({ poolId });
+    if (!currentPoolData || !currentPoolData.pool) {
+      throw new Error("Pool data is not found or is invalid.");
+    }
+    pool = currentPoolData.pool;
+  }
+
+  if (!pool || !pool.tokens) {
+    return 0;
+  }
```
ribeirojose, line 361: Not sure what's the use case that makes us need this here (and I ask myself if we would've needed it elsewhere) but you'd help your reader if you encapsulate this in another function:

```
async function fetchPoolData(chain: string, poolId: string, timeStart: number, timeEnd: number): Promise<Pool | null> {
  const rangeData = await poolsWithCache.gql(chain).poolSnapshotInRange({
    poolId,
    timestamp: generateDateRange(timeStart, timeEnd),
  });
  if (rangeData.poolSnapshots.length > 0) {
    return rangeData.poolSnapshots[0].pool;
  }
  const currentPoolData = await pools.gql(chain).Pool({ poolId });
  return currentPoolData?.pool || null;
}

// ... then in calculateTokenApr

...
  const pool = await fetchPoolData(chain, poolId, timeStart, timeEnd);
  if (!pool || !pool.tokens) return 0;
...
```

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+  ) {
+    const isExemptFromYieldProtocolFee = pool?.tokens.find(
+      ({ address }) => address === tokenAddress,
+    )?.isExemptFromYieldProtocolFee;
+
+    if (isExemptFromYieldProtocolFee) {
+      apr = tokenYield;
+    } else {
+      //source: https://github.com/balancer/balancer-sdk/blob/f4879f06289c6f5f9766ead1835f4f4b096ed7dd/balancer-js/src/modules/pools/apr/apr.ts#L134
+      apr = tokenYield * (1 - pool.protocolYieldFeeCache || 0.5);
```
ribeirojose, line 378: can we assign this random 0.5 to a constant like `DEFAULT_PROTOCOL_YIELD_FEE`?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+
+  if (rangeData.poolSnapshots.length > 0) {
+    poolData = rangeData.poolSnapshots[0].pool;
+  } else {
+    const currentPoolData = await pools.gql(poolNetwork).Pool({ poolId });
+    if (!currentPoolData || !currentPoolData.pool) {
+      throw new Error("Pool data is not found or is invalid.");
+    }
+    poolData = currentPoolData.pool;
+  }
```
ribeirojose, line 414: this is the same code we had above, right? 😉

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+
+  const poolTokenData = poolData.tokens?.filter((token) => {
+    return token.address !== poolData.address;
+  });
+
+  if (!Array.isArray(poolTokenData)) {
```
ribeirojose, line 420: shouldn't this at least log a warning as well? when would this not be an array?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+    const currentPoolData = await pools.gql(poolNetwork).Pool({ poolId });
+    if (!currentPoolData || !currentPoolData.pool) {
+      throw new Error("Pool data is not found or is invalid.");
+    }
+    poolData = currentPoolData.pool;
+  }
+
+  const poolTokenData = poolData.tokens?.filter((token) => {
+    return token.address !== poolData.address;
+  });
```
ribeirojose, line 418: I think you mean that this is the pool tokens filtering out its own BPT, right? Think your names could help understand more why this is going on here. Perhaps we should also add a comment explaining why ignore it.
```suggestion
    const relevantTokens = poolData.tokens?.filter(token => token.address !== poolData.address) || [];
```

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+
+  if (!Array.isArray(poolTokenData)) {
+    return 0;
+  }
+
+  const token = poolTokenData.find((t) => t.address === tokenAddress);
+  if (!token) {
+    // eslint-disable-next-line no-console
```
ribeirojose, line 426: could've disabled this for the whole file.

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+
+  const tokensPrices = await Promise.all(
+    poolTokenData.map(async (token) => {
+      if (
+        vunerabilityAffecteRateProviders.some(
+          ({ address }) =>
+            address.toLowerCase() === token.address.toLowerCase(),
+        )
+      ) {
+        return 1;
```
ribeirojose, line 443: why is 1 the default?

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+  }, 0);
+
+  const tokenIdx = poolTokenData.findIndex((t) => t.address === tokenAddress);
+  const tokenPrice = tokensPrices[tokenIdx];
+
+  const tokenValue = tokenPrice * parseFloat(token.balance);
+  const tokenWeight = tokenValue / totalValue;
+
+  return tokenWeight;
+}
```
ribeirojose, line 479: again a bit hard to parse/understand. I'd aim for something like this:

```
async function getTokenWeight(
  tokenAddress: string, 
  startAtTimestamp: number, 
  endAtTimestamp: number, 
  poolNetwork: string, 
  poolId: string
): Promise<number> {
  const { tokens, poolType, address } = await fetchPoolData(poolNetwork, poolId, startAtTimestamp, endAtTimestamp);
  
  const relevantTokens = tokens?.filter(({ address }) => address !== address) || [];
  const targetToken = relevantTokens.find(({ address }) => address === tokenAddress);
  
  if (!targetToken) {
    console.warn(`Token not found: ${tokenAddress}`);
    return 0;
  }
  
  if (poolType === 'WEIGHTED' && targetToken.weight) return parseFloat(targetToken.weight);
  
  const tokenPrices = await fetchTokenPrices(relevantTokens, endAtTimestamp, poolNetwork);
  const totalValue = relevantTokens
    .map(({ balance }, idx) => parseFloat(balance) * tokenPrices[idx])
    .filter(val => !isNaN(val))
    .reduce((acc, val) => acc + val, 0);
  
  const tokenIdx = relevantTokens.findIndex(({ address }) => address === tokenAddress);
  return (parseFloat(targetToken.balance) * tokenPrices[tokenIdx]) / totalValue;
}
```

ribeirojose: also, since you're fetching the pool data from both functions, you could just fetch it beforehand in `getAPRFromRateProviderInterval` and pass it as an argument. Will save us another round trip to the subgraph =)

ribeirojose (commented):
Great findings and explanation in the video so far -- thanks for the effort you put here! 

I'm not approving this even though I've seen us merging things w/o approval before (and I know the risk that this will get merged anyways). But I think we can take some more time to focus on some code quality issues here. I left a few comments, let me know if they make sense to you

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
luizakp: Agree! During development, I just like to make sure I'm not getting errors because I swap the props order xD but will keep it consistent 🫡

luizakp:
> Great findings and explanation in the video so far -- thanks for the effort you put here!
> 
> I'm not approving this even though I've seen us merging things w/o approval before (and I know the risk that this will get merged anyways). But I think we can take some more time to focus on some code quality issues here. I left a few comments, let me know if they make sense to you

Sure, focusing on making the code better is a good idea! I just wanted to quickly show you what I have done to hear what you think. 😁 I know the code is not very well organized, but I wanted to share it with the team and also send it to Balancer soon

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
luizakp: actually, I was thinking about this, how not to blow the token weight when one of the tokens doesn't have their price on coinllama

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
luizakp: I was getting some type warnings, but this is not needed anymore hehehe

apps/balancer-tools/src/app/apr/(utils)/tokenApr.ts
```diff
@@ -285,3 +321,159 @@ const getImplementationAddress = async (
...
+    pool.poolType === PoolTypeEnum.META_STABLE ||
+    pool.poolType?.includes("Gyro")
+  ) {
+    apr = tokenYield * (pool.protocolYieldFeeCache || 0.5);
+  }
+
+  apr = (apr || 0) * tokenWeight;
+
+  return apr;
+}
```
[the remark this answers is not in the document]
luizakp: Thank you for this! Not used to `class`. But I think the outcome is much better! I also broke into files because this `tokenApr.ts` was getting too big and hard to understand

apps/balancer-tools/src/app/apr/(utils)/tokenApr/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+): Promise<number[]> {
+  return Promise.all(
+    relevantTokens.map(async (token) => {
+      if (
+        vunerabilityAffecteRateProviders.some(
+          ({ address }) =>
+            address.toLowerCase() === token.address.toLowerCase(),
+        )
+      ) {
+        return 1;
```
ribeirojose, line 83: Why not return 0 if this is affected by the vulnerability? 1 looks pretty misleading to me.

ribeirojose (dismissed):

apps/balancer-tools/src/app/apr/(utils)/tokenApr/utils.ts
luizakp: you are right! in this case it should be 0

ribeirojose: coming back to this, and it should also consider if the endTimestamp is AFTER the vulnerability right?

apps/balancer-tools/src/app/apr/(utils)/tokenApr/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+          endAtTimestamp,
+          token.address,
+          parseInt(poolNetwork),
+        );
+
+        if (tokenPrice === undefined) {
+          console.warn(
+            `Failed fetching price for ${token.symbol} (network: ${poolNetwork}, addr: ${token.address}) at ${endAtTimestamp}`,
+          );
+          return 1;
```
ribeirojose, line 95: also would add a TODO to review this -- this is tech debt to me. we don't need to solve it now but it needs to be registered. same as with the above

luizakp: here: https://linear.app/bleu-llc/issue/BAL-1164/default-price-when-fails-to-fetch-token-price

PR #447, opened 2023-10-26
[Milkman] app setup

luizakp:
https://www.loom.com/share/b67a092d20da497fbf22c69bbecfd35f?sid=26c3b6e3-3313-47c9-ad2b-0b6576fbf8e3


apps/cow-tools/src/lib/gql/server.ts
luizakp, line 1: @ribeirojose any suggestion to make something similar to codegen?

luizakp: suggestion here: https://github.com/bleu-fi/balancer-tools/pull/447/commits/d95509f882c5c0f4c1983daf06033da60c472663

apps/cow-tools/next.config.js
```diff
@@ -0,0 +1,105 @@
+/* eslint-env node */
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { withSentryConfig } = require("@sentry/nextjs");
+
+/** @type {import('next').NextConfig} */
+const moduleExports = {
```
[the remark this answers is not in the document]
luizakp: @ribeirojose I used the same config from balancer-tools. Why was Hound caught here but not there? it's because these files were added there before Hound?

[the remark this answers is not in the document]
ribeirojose: Not sure, but you can ignore it

apps/cow-tools/src/__tests__/components/Button.test.tsx
```diff
@@ -0,0 +1,78 @@
+import { DotIcon } from "@radix-ui/react-icons";
```
ribeirojose, line 1: do need these tests here?

ribeirojose (dismissed):
LGTM, but I'd suggest we remove whatever component is not being used and just copy those that are being used. The current approach makes the whole thing so complex already

PR #448, opened 2023-10-27
[APR] fix loading more and filter by pool types

luizakp:
This PR:
- Fix more pools load button
- remove pools that have bb-pools as tokens
- fix pool type filter

https://www.loom.com/share/bd1ac72dab39485fac0bbae122547c16?sid=383ed14e-c576-4a38-8f83-8ab214430a12


packages/gql/src/balancer/pools.ts
```diff
@@ -195,12 +195,14 @@ export const aprPools = gql`
...
     $maxTvl: BigDecimal
     $block: Int
+    $tokensAsVunerabilityAffectedPools: [String!]
   ) {
     pools(
       where: {
         createTime_lte: $createdBefore
         totalLiquidity_gte: $minTvl
         totalLiquidity_lte: $maxTvl
+        tokens_: { address_not_in: $tokensAsVunerabilityAffectedPools }
```
ribeirojose, line 205: So we're not going to report this even historically?

ribeirojose (approved):
Lgtm, just left one question

packages/gql/src/balancer/pools.ts
luizakp: 🫠 you are right

luizakp: added a tech debt here: https://linear.app/bleu-llc/issue/BAL-1165/handle-bb-pools-proportion-for-yield-apr 
talked to Fabio and the majority of bb pools have token yield, and the token apr will never be calculated without the price/proportion

So will filter out those pools and add them back into this task

PR #456, opened 2023-11-01
chore(apr): use postgres instead of redis

ribeirojose:


luizakp:
we will use the branch `dev-db-apr`

PR #462, opened 2023-11-03
chore(milkman): use factory/strategy pattern for transactions

ribeirojose:


apps/cow-tools/src/lib/transactionFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+    tokenAddressToSell,
+    tokenAddressToBuy,
+    toAddress,
+    amount,
+    priceChecker,
+    args,
+  }: MilkmanOrderArgs): BaseTransaction {
+    const priceCheckerInfo = priceCheckerInfoMapping[priceChecker];
+    const priceCheckerAddress = priceCheckerInfo.addresses[goerli.id];
+    const priceCheckerData = encodePriceCheckerData(priceChecker, args);
```
ribeirojose, line 76: I wasn't sure this should be `[args]` as before because this was already being passed as an array to this function. I couldn't run a transaction locally to test =/

yvesfracari (approved):
Thanks! Nice ideas

PR #471, opened 2023-11-09
Add new order form validations pt 1

yvesfracari:
https://www.loom.com/share/aa790d3a176948a3abd840cbf3d19bd2

### DIFF
--- a/apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx
+++ b/apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx
@@ -477,7 +477,10 @@ function FormSelectPriceChecker({
             {...register(arg.name)}
           />
         ))}
-      <FormFooter transactionStatus={TransactionStatus.ORDER_STRATEGY} />
+      <FormFooter
+        transactionStatus={TransactionStatus.ORDER_STRATEGY}
+        disabled={!selectedPriceChecker}
+      />
     </Form>
   );
 }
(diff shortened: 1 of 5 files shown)


apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx
```diff
@@ -98,57 +94,9 @@ function TransactionCard({
...
-        toAddress: data.receiverAddress,
-        amount: sellAmountBigInt,
-        priceChecker: data.priceChecker,
-        args: priceCheckersArgs,
-      },
-    ]);
-    router.push(`/milkman/${network}`);
-  }
+  const [orderOverviewData, setOrderOverviewData] = useState<FieldValues>();
+  const [priceCheckerData, setPriceCheckerData] = useState<FieldValues>();
```
Comment (luizakp) [inline apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx:99]: Think I wouldn't create a state form this... because one of the points of using react hook form is to deal with the state within the form

Comment (luizakp) [reply apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx:99]: we are using this to combine the form data, correct?

Comment (luizakp) [reply apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx:99]: a blocker here is the priceChecker form changing for each price checker

apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx
```diff
@@ -252,86 +211,160 @@ function FormHeader({
...
+    return assets.find(
+      (tokenBalance) => tokenBalance.tokenInfo.address === address,
+    );
+  }
+
+  const [isValidFromNeeded, setIsValidFromNeeded] = useState(
+    !!defaultValues?.validFrom,
   );
+  const [tokenBuy, setTokenBuy] = useState<TokenBalance | undefined>();
+  const [tokenSell, setTokenSell] = useState<TokenBalance | undefined>();
```
Comment (luizakp) [inline apps/cow-tools/src/app/milkman/[network]/order/new/page.tsx:253]: would remove this

suggestion here: https://github.com/bleu-fi/monorepo/pull/473

PR #480, opened 2023-11-14
[MilkmApp] Add chainlink price checker

yvesfracari:
https://www.loom.com/share/f395384f9b54473a9fb6c7d4969a3fe4?sid=0514dd82-157f-4d2b-a5ab-27b0071adaad


apps/cow-tools/src/lib/priceCheckers.ts
```diff
@@ -72,72 +83,125 @@ export const priceCheckerInfoMapping = {
...
+        label: "Revert price feeds",
+        inputType: "checkbox",
+        toExpectedOutCalculator: false,
+        convertInput: (input: boolean) => input,
+      },
+    ] as PriceCheckerArgument[],
+    name: PRICE_CHECKERS.CHAINLINK,
+    hasExpectedOutCalculator: true,
+    getSchema: getChainlinkSchema,
+  },
```
ribeirojose, line 124: 👌

apps/cow-tools/src/app/milkman/utils/type.ts
```diff
@@ -1,7 +1,7 @@
 export enum TransactionStatus {
   ORDER_OVERVIEW = "Order overview",
   ORDER_STRATEGY = "Order strategy",
-  ORDER_RESUME = "Order resume",
+  ORDER_SUMMARY = "Order resume",
```
ribeirojose, line 4:
```suggestion
  ORDER_SUMMARY = "Order summary",
```

apps/cow-tools/src/app/milkman/(components)/orderForm/Resume.tsx
```diff
@@ -87,38 +87,59 @@ export function OrderResume({
...
+                  </span>
+                )
+              );
+            case "addressesPriceFeeds":
+              return (
+                value && (
+                  <span key={field.key}>
+                    {field.label}:{" "}
+                    {value.map(
+                      (address: string) => truncateAddress(address)?.toString(),
```
ribeirojose, line 115: why call toString here?

apps/cow-tools/src/app/milkman/(components)/orderForm/Resume.tsx
ribeirojose, line 1: `apps/cow-tools/src/app/milkman/(components)/orderForm/Summary.tsx`

apps/cow-tools/src/app/milkman/(components)/orderForm/Resume.tsx
```diff
@@ -87,38 +87,59 @@ export function OrderResume({
...
-          if (field.key === "tokenBuy") {
-            return (
-              value && (
-                <span key={field.key}>
-                  {field.label}: {value.symbol} (
-                  {truncateAddress(data.tokenBuy.address)})
-                </span>
-              )
-            );
+          switch (field.key) {
```
ribeirojose, line 90: these switch case statements within components render make it so hard to reason about the whole component logic -- I trust your code here works but `would` make an effort to modularize a bit more this code and avoid inline logic, perhaps you could start taking a read at code reviews in other projects as well to see how other folks have handled this

apps/cow-tools/src/app/milkman/(components)/orderForm/PriceChecker.tsx
```diff
@@ -92,22 +100,297 @@ export function FormSelectPriceChecker({
...
+      setValue("priceFeedAddress", priceFeed.address);
+      setValue("reversed", priceFeed.reversed);
+    }
+  }, []);
+
+  const newPriceFeedForm = useForm<typeof priceFeedChainlinkSchema._type>({
+    resolver: zodResolver(priceFeedChainlinkSchema),
+  });
+  const { register, watch, setValue } = newPriceFeedForm;
+  const formData = watch();
```
ribeirojose, line 310: couldn't we just use the parent form instead of adding another one here?

ribeirojose (approved):

luizakp:
🔝 just one question did you test the transaction with more than one oracle to check if it goes through?

apps/cow-tools/src/app/milkman/(components)/orderForm/PriceChecker.tsx
yvesfracari: I divided it into two forms because it would be easier to test if each price feed is the contract that we expect here. Otherwise, we would have to deal with some errors on the other component. Maybe we can take a look at it together later, what do you think about it?

yvesfracari:
I will test all of them today!

PR #481, opened 2023-11-14
[APR] Fees apr

luizakp:
only API working without limit, order or sort


apps/balancer-tools/src/app/apr/(utils)/getBALPriceForDateRange.ts
```diff
@@ -1,52 +1,43 @@
...
+        price_usd
+       FROM
+        token_prices
+       WHERE
+       token_address = '${sql.raw(BAL_TOKEN_ADDRESS)}'
+       AND timestamp >= '${sql.raw(startAt.toISOString())}'
+       AND timestamp <= '${sql.raw(endAt.toISOString())}'
+      ) AS subquery
+  `);
+  //TODO when there is no price on db
```
luizakp, line 32: thinking about using the old version as a fallback and insert the result on the db

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -25,54 +23,79 @@ export interface calculatePoolData extends Omit<PoolStatsData, "apr"> {
...
-    poolId,
-  });
-
-  let chosenData = res.poolSnapshots[0];
-
-  if (res.poolSnapshots.length == 0) {
-    const retryGQL = await fetchPoolSnapshots({
-      to: to - SECONDS_IN_DAY,
-      from: from - SECONDS_IN_DAY * 7,
+export async function fetchPoolData(poolId: string, to: number) {
```
luizakp, line 26: this fails if there is no snapshot for that day

ribeirojose: ok, so we need to update the snapshot table rigiht?

apps/balancer-tools/src/app/apr/(utils)/getBALPriceForDateRange.ts
ribeirojose: lgtm

apps/balancer-tools/src/db/migrations/0000_striped_hiroim.sql
```diff
@@ -64,8 +64,6 @@ CREATE TABLE IF NOT EXISTS "pool_snapshots" (
 	"total_shares" numeric,
```
ribeirojose, line 64: generate new migrations instead of deleting this one, otherwise you'll force everyone to drop their databases

apps/balancer-tools/src/app/apr/api/(utils)/fetchForDateRange.ts
```diff
@@ -1,142 +1,67 @@
...
-  const allFetchedPools = await Promise.all(
-    networks.map(
-      async (network) =>
-        await fetchPoolsFromNetwork(network, endDate, maxTvl, minTvl),
-    ),
-  );
+  maxTvl: number = 10_000_000_000,
+  minTvl: number = 10_000,
+): Promise<{ [key: string]: PoolStatsData[] }> {
+  const existingPoolForDate = await db.execute(sql`
```
ribeirojose, line 19: this was an easy one to use drizzle's own ORM functions to build :grimacing:

apps/balancer-tools/src/app/apr/api/(utils)/fetchForDateRange.ts
```diff
@@ -1,142 +1,67 @@
...
-    const perDayData: { [key: string]: PoolStatsData[] } = {};
+  await Promise.all(
+    existingPoolForDate.map(async (pool) => {
+      let gaugesData;
+      try {
+        gaugesData = await fetcher<PoolStatsResults>(
+          `${BASE_URL}/apr/api?startAt=${formatDateToMMDDYYYY(
+            startDate,
+          )}&endAt=${formatDateToMMDDYYYY(endDate)}&poolId=${pool.id}`,
+        );
```
ribeirojose, line 44: we could theoretically get rid of this now, and it also skews our results a bit given our new reality. have you discussed this with Fábio?

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -25,54 +23,79 @@ export interface calculatePoolData extends Omit<PoolStatsData, "apr"> {
...
+    pt.weight AS weight,
+    t.symbol AS symbol
+  FROM
+    pool_tokens pt
+    JOIN tokens t ON t.address = pt.token_address
+  WHERE
+  pool_external_id = '${sql.raw(poolId)}'
+  ORDER BY
+    pt.weight DESC;
+  `);
```
ribeirojose, line 89: prefer drizzle queries since they'll return you with a typed result: https://orm.drizzle.team/docs/select

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
```diff
@@ -25,54 +23,79 @@ export interface calculatePoolData extends Omit<PoolStatsData, "apr"> {
...
-    // if the retry was needed then the volume on that day is 0
-    chosenData.swapVolume = 0;
-  }
-
-  const liquidity = Number(chosenData.liquidity);
-
-  const volume = Number(chosenData.swapVolume);
+      CAST(liquidity AS NUMERIC) AS liquidity,
+      CAST(swapVolume AS NUMERIC) AS volume,
+      (CAST(liquidity AS NUMERIC) / CAST(totalShares AS NUMERIC)) AS bpt_price
```
ribeirojose, line 50: we should probably already insert these numbers in the DB as numeric, wdyt?

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
```diff
@@ -133,95 +105,67 @@ export async function calculateAPRForDateRange(
...
+      MAX(ps.timestamp) OVER w AS end_timestamp
+    FROM
+      pool_snapshots ps
+      JOIN pools p ON p.external_id = ps.pool_external_id
+    WHERE
+      ps.pool_external_id = '${sql.raw(poolId)}'
+      AND ps.timestamp BETWEEN '${sql.raw(
+        epochToDate(from).toISOString(),
+      )}' AND '${sql.raw(epochToDate(to).toISOString())}'
+    WINDOW w AS (ORDER BY ps.timestamp RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)
```
ribeirojose, line 149: 😅 I know I'm responsible for this monster

ribeirojose (approved):

apps/balancer-tools/src/app/apr/(utils)/calculatePoolStats.ts
luizakp: now snapshot table has all day since balancer start

apps/balancer-tools/src/app/apr/api/(utils)/fetchForDateRange.ts
luizakp: Yes! we discussed and we think it makes more sense to show the average of the days instead of the apr of the range date. What we have to do is to make that more clear to the user

apps/balancer-tools/src/app/apr/(utils)/calculateApr.ts
luizakp: we worked on this over the weekend, no need to this monster anymore ;)

apps/balancer-tools/src/app/apr/api/(utils)/fetchForDateRange.ts
luizakp: AA I think now I got what you mean, we don't have the add one layer of API, we can just call the query directly

ribeirojose (approved):

PR #494, opened 2023-11-24
[APR] veBAL APR on DB

luizakp:
- create rounds table
- create emissions table
- populate gaugeSnapshot
- calculate veBAL APR

### DIFF
--- a/apps/balancer-tools/seed.ts
+++ b/apps/balancer-tools/seed.ts
@@ -17,7 +17,6 @@ import {
   poolSnapshots,
   poolSnapshotsTemp,
   tokenPrices,
-  vebalApr,
   vebalRounds,
 } from "#/db/schema";
 import * as balEmissions from "#/lib/balancer/emissions";
@@ -642,6 +641,7 @@ async function seedBalEmission() {
 
 async function calculateApr() {
   // Fee APR
+  logIfVerbose("Seeding Fee APR");
   await db.execute(sql`
   INSERT INTO swap_fee_apr (timestamp, pool_external_id, collected_fees_usd, value, external_id)
   SELECT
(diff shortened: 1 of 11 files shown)


apps/balancer-tools/seed.ts
```diff
@@ -262,6 +268,81 @@ async function extractGauges() {
   await processGauges(response.data);
 }
 
+async function extractGaugesSnapshot() {
```
luizakp, line 271: This gets the relative weight of each gauge every day since the gauge was created. This is getting one by one, so no need to worry about rate limit. Also, this can be a multicall to be faster, but definitely not a must

apps/balancer-tools/seed.ts
```diff
@@ -334,24 +415,6 @@ ON
...
-    CASE 
-        WHEN p1.liquidity = 0 THEN 0
-        ELSE ((p1.swap_fees - p2.swap_fees) * (1 - COALESCE(p1.protocol_swap_fee_cache, 0.5)) / p1.liquidity) * 365 * 100 
-    END AS "value",
-  p1.external_id
-FROM pool_snapshots p1
-LEFT JOIN pool_snapshots p2 
-ON p1. "timestamp" = p2. "timestamp" + INTERVAL '1 day'
-AND p1.pool_external_id = p2.pool_external_id
-ON CONFLICT (external_id) DO NOTHING;
```
luizakp, line 354: moved this to a proper apr function

apps/balancer-tools/seed.ts
```diff
@@ -523,83 +592,98 @@ DO UPDATE SET chain_id = EXCLUDED.chain_id, updated_at = NOW();
 `);
 }
 
-// async function fetchTokenPrices() {
```
luizakp, line 526: removed this for now as this has to be refactor to work properly

apps/balancer-tools/seed.ts
```diff
@@ -523,83 +592,98 @@ DO UPDATE SET chain_id = EXCLUDED.chain_id, updated_at = NOW();
...
+          weekEmission: String(weeklyBalEmission),
+        })
+        .onConflictDoNothing()
+        .execute();
+    } catch (error) {
+      console.error(`${error}`);
+    }
+  }
+
+  // balEmissions.weekly(endAtTimestamp);
```
luizakp, line 639:
```suggestion

```

apps/balancer-tools/seed.ts
```diff
@@ -523,83 +592,98 @@ DO UPDATE SET chain_id = EXCLUDED.chain_id, updated_at = NOW();
...
+
+async function seedBalEmission() {
+  const timestamps = await db
+    .selectDistinct({ timestamp: poolSnapshots.timestamp })
+    .from(poolSnapshots);
+
+  for (const { timestamp } of timestamps) {
+    if (!timestamp) continue;
+    try {
+      const weeklyBalEmission = balEmissions.weekly(dateToEpoch(timestamp));
```
luizakp, line 624: @mendesfabio used to logic we already have on JS to not take so much time to get the query right

apps/balancer-tools/seed.ts
```diff
@@ -523,83 +592,98 @@ DO UPDATE SET chain_id = EXCLUDED.chain_id, updated_at = NOW();
...
+FROM pool_snapshots ps
+LEFT JOIN vebal_rounds vr ON ps.timestamp BETWEEN vr.start_date AND vr.end_date
+JOIN gauges g ON g.pool_external_id = ps.pool_external_id
+JOIN gauge_snapshots gs ON g.address = gs.gauge_address
+    AND vr.round_number = gs.round_number
+LEFT JOIN token_prices tp ON tp.token_address = '0xba100000625a3754423978a60c9317c58a424e3d'
+    AND tp.timestamp = ps.timestamp
+LEFT JOIN bal_emission be ON be.timestamp = ps.timestamp
+ON CONFLICT (external_id) DO NOTHING;
+`);
```
luizakp, line 684: veBAL APR will always be a little different from what is showing on Balancer's UI because the only consider the staked tokens and we consider the whole liquidity @ribeirojose 

I discussed this with @mendesfabio  and the gaugeTotalSupply (the staked tvl) is not saved by timestamp in the subgraph, that's why we are using this approximation. We concluded the values are accurate if they are within the stacking APR range, closer to the min staked APR

apps/balancer-tools/seed.ts
```diff
@@ -658,12 +742,14 @@ async function runETLs() {
...
   await seedNetworks();
+  await seedVebalRounds();
   await ETLPools();
   await ETLSnapshots();
   await ETLGauges();
-  // await fetchTokenPrices(); -> this is not necessary for every pool, just for the ones that haev token rewards and token yield
+  await seedBalEmission();
   await fetchBalPrices();
-  // await fetchBlocks();
+  await ETLGaugesSnapshot();
```
luizakp, line 751: this takes forever, that's why is the last before the apr calculation

apps/balancer-tools/src/app/apr/(utils)/getRelativeWeight.ts
```diff
@@ -36,33 +23,18 @@ async function readContract(gaugeAddress: Address, time: number) {
...
-        gaugeAddress: gaugeAddress,
-        timestamp: new Date(time),
-        relativeWeight: String(Number(data) / 1e18),
-      })
-      .onConflictDoNothing()
-      .returning();
-
-    return gaugeSnapshot.relativeWeight
-      ? gaugeSnapshot.relativeWeight / 1e18
-      : 0;
```
luizakp, line 62: we deal with this in `seed.ts`

apps/balancer-tools/seed.ts
```diff
@@ -453,6 +516,8 @@ async function transformGauges() {
   DO UPDATE SET is_killed = EXCLUDED.is_killed
   WHERE gauges.address IS DISTINCT FROM EXCLUDED.address;`);
 
+  //TODO: get raw_data to the gauges table, this is creating new rows and deleting old ones, so the id starts on 267
```
ribeirojose, line 519: ❓

apps/balancer-tools/seed.ts
ribeirojose: if it has to be refactored, why remove it? Is not of that usable? 😅

apps/balancer-tools/seed.ts
```diff
@@ -523,83 +592,98 @@ DO UPDATE SET chain_id = EXCLUDED.chain_id, updated_at = NOW();
...
+        ELSE (52 * (be.week_emission * gs.relative_weight * tp.price_usd) / ps.liquidity) * 100
+    END AS value,
+    ps.pool_external_id,
+    ps.external_id
+FROM pool_snapshots ps
+LEFT JOIN vebal_rounds vr ON ps.timestamp BETWEEN vr.start_date AND vr.end_date
+JOIN gauges g ON g.pool_external_id = ps.pool_external_id
+JOIN gauge_snapshots gs ON g.address = gs.gauge_address
+    AND vr.round_number = gs.round_number
+LEFT JOIN token_prices tp ON tp.token_address = '0xba100000625a3754423978a60c9317c58a424e3d'
```
ribeirojose, line 680: 👀

apps/balancer-tools/seed.ts
ribeirojose: is this idempotent?

apps/balancer-tools/src/db/migrations/0000_wise_justice.sql
ribeirojose, line 1: changing the base migration again :/

ribeirojose (approved):
lgtm

PR #506, opened 2023-12-01
[APR] refactor pool table sort and filtering using db

luizakp:
https://www.loom.com/share/dac98b32cef0468681803b5f29540d89?sid=db762227-02e6-4e65-8a0c-328057caa44c


luizakp:
@ribeirojose you can see in the video that it takes a while to return on the frontend, that's because next route.push is a promise and it waits to be resolved to then push, but our index page has the suspense to let the user know it's loading... I already google a bit, but I will break into another task to not let this one blocked 

https://linear.app/bleu-llc/issue/BAL-1194/redirect-without-waiting-promise-to-resolve-or-add-loader

luizakp:
> @luizakp just wanted to validate one other thing: is this supposed to be merged to `main`? Asking because I know you were doing these APR updates in another branch

now yes! that branch was merged, I'll clean the branches

PR #507, opened 2023-12-04
[APR] Yield apr on db

luizakp:


apps/balancer-tools/src/app/apr/(utils)/fetchForDateRange.ts
```diff
@@ -60,27 +61,43 @@ export async function fetchDataForDateRange({
...
+        sql<number>`cast(sum(${swapFeeApr.value}) /  count(${poolSnapshots.timestamp})  as decimal)`.as(
+          "avgFeeApr",
+        ),
+      avgVebalApr:
+        sql<number>`cast(sum(${vebalApr.value}) /  count(${poolSnapshots.timestamp})  as decimal)`.as(
+          "avgVebalApr",
+        ),
       avgVolume:
-        sql<number>`cast(avg(${poolSnapshots.swapVolume}) as decimal)`.as(
+        sql<number>`cast(sum(${poolSnapshots.swapVolume}) /  count(${poolSnapshots.timestamp})  as decimal)`.as(
```
ribeirojose, line 90: This query had been repeated in the file above, no?

packages/balancer-apr/package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "@vercel/postgres": "^0.5.1",
     "drizzle-orm": "^0.29.1",
+    "p-throttle": "^6.0.0",
```
ribeirojose, line 40: Why?

packages/balancer-apr/src/db/migrations/0006_young_magik.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE "pool_rate_providers" ADD COLUMN "external_id" varchar;--> statement-breakpoint
```
ribeirojose, line 1: 👍 good call

packages/balancer-apr/src/db/migrations/0007_certain_jimmy_woo.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "blocks" ADD COLUMN "external_id" varchar;--> statement-breakpoint
```
ribeirojose, line 1: What's the external id of a block?

apps/balancer-tools/src/app/apr/(components)/PoolListTable.tsx
```diff
@@ -141,27 +142,6 @@ export function PoolListTable({
...
-                    <Button
-                      className="py-3 sticky left-0 sm:w-full flex content-center justify-center gap-x-3 rounded-t-none rounded-b disabled:cursor-not-allowed"
-                      shade="medium"
-                      // disabled={isLoadingMore || !hasMorePools}
-                      disabled={true}
-                    >
-                      {/* {isLoadingMore ? (
-                        <Spinner size="sm" />
-                      ) : hasMorePools ? (
-                        <> */}
```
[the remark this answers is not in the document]
luizakp: https://linear.app/bleu-llc/issue/BAL-1191/loading-more-pools-button

apps/balancer-tools/src/app/apr/(utils)/fetchAvgDataForPoolIdDateRange.ts
```diff
@@ -11,13 +12,26 @@ export async function fetchAvgDataForPoolIdDateRange(
...
+    .groupBy(yieldTokenApr.poolExternalId, yieldTokenApr.timestamp)
+    .as("yieldTokenAprSum");
   const poolStatsData = await db
     .select({
       poolExternalId: swapFeeApr.poolExternalId,
-      avgApr: sql<number>`cast(avg(   
-        coalesce(${swapFeeApr.value},0) + coalesce(${vebalApr.value},0)
-      ) as decimal)`,
-      avgLiquidity: sql<number>`cast(avg(${poolSnapshots.liquidity}) as decimal)`,
+      avgApr:
```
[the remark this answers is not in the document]
luizakp: because it does the avg of the values that it finds.

Example: for 4 dates if only one day has veBAL APR and that value is 8, then the avg should be 8/4 = 2, but postgres avg is returning 8, as the other days don't exist on veBAL APR table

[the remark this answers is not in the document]
ribeirojose: but why are the other ones missing from the table? 👀 I thought this was the issue we solved originally by populating pool snapshots everyday, even on missing days?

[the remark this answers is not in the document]
luizakp: yup this for snapshots table, but for apr tables I don't think that makes much sense because we would have a much of rows with 0 APR

apps/balancer-tools/src/app/apr/(utils)/fetchForDateRange.ts
luizakp: yes, they are meant for two different occasion

packages/balancer-apr/package.json
luizakp: I was getting rate limited by the contracts even with multicall configuration `batchSize` and `wait`. At this point prefered to over kill and get all the data in instead of investing more time debugging it

apps/balancer-tools/src/app/apr/(utils)/fetchAvgDataForPoolIdDateRange.ts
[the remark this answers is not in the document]
ribeirojose: if it'd make things easier that's ok -- a row costs marginally 0 to us

PR #547, opened 2023-12-22
[Preferential gauge] Frontend PoC

luizakp:


packages/gql/src/balancer-gauges/gauge.ts
```diff
@@ -7,3 +7,34 @@ export const gaugeSymbol = gql`
     }
   }
 `;
+
+export const poolsWithGauge = gql`
+  query PreferentialGauge {
+    pools(first: 1000, where: { gauges_: { gauge_not: null } }) {
```
mendesfabio, line 13:
```suggestion
    pools(first: 1000, where: { gauges_not: null }) {
```

packages/gql/src/balancer-gauges/gauge.ts
```diff
@@ -7,3 +7,34 @@ export const gaugeSymbol = gql`
...
+      gauges {
+        id
+      }
+    }
+  }
+`;
+
+export const gaugeInfo = gql`
+  query PoolPreferentialGauge($poolId: Bytes!) {
+    pools(where: { poolId_in: [$poolId], gauges_: { gauge_not: null } }) {
```
mendesfabio, line 24:
```suggestion
    pools(where: { poolId_in: [$poolId], gauges_not: null }) {
```

packages/gql/src/balancer-gauges/gauge.ts
```diff
@@ -7,3 +7,34 @@ export const gaugeSymbol = gql`
...
+        id
+      }
+      gauges {
+        gauge {
+          liquidityGauge {
+            id
+            symbol
+          }
+        }
+      }
```
mendesfabio, line 37:
```suggestion
      gauges {
            id
            symbol
      }
```

packages/gql/src/balancer-gauges/gauge.ts
mendesfabio: I believe this is already enough - naming here doesn't help much but we need pools which have liquidity gauges (so it can't be null) but it's fine if the gauge attribute is empty - it simply indicates whether you can vote on that gauge, but that's not what we're looking for here.

packages/gql/src/balancer-gauges/gauge.ts
mendesfabio: same explanation from above

packages/gql/src/balancer-gauges/gauge.ts
mendesfabio: feel free to ignore if it causes much trouble down the line, but `gauges` already has `id` and `symbol` and you can use it

mendesfabio (dismissed):

ribeirojose:
Do you need any help fixing the conflicts and merging this @luizakp ?

luizakp:
> Do you need any help fixing the conflicts and merging this @luizakp ?

ops forgot I can merge this 😅

ribeirojose:
> > Do you need any help fixing the conflicts and merging this @luizakp ?
> 
> ops forgot I can merge this 😅

😬

yvesfracari:
Are we still planning to merge this? If not we can just close the PR. @luizakp

PR #607, opened 2024-02-28
[CoW Amm Deployer] Add stop AMM feature

yvesfracari:


apps/cow-amm-deployer/src/app/amm/HomePageWrapper.tsx
```diff
@@ -114,7 +119,14 @@ export function HomePageWrapper() {
             <Button
               className="flex items-center gap-1 py-3 px-6 "
               color="tomato"
-              disabled
+              onClick={async () => {
+                await sendTransactions([
+                  {
+                    type: TRANSACTION_TYPES.STOP_COW_AMM,
```
ribeirojose, line 125: 🙆‍♂️ cool! I'm assuming all this had already been implemented then?

apps/cow-amm-deployer/src/lib/transactionFactory.ts
```diff
@@ -140,11 +145,25 @@ class CowAmmCreateTx implements ITransaction<creteCowAmmArgs> {
     };
   }
 }
+
+class CowAmmStopTx implements ITransaction<stopCowAmmArgs> {
```
ribeirojose, line 149: any comments on why use `Stop` when the fn name is `Close`? Do you think that's more friendly/meaningful to users?

ribeirojose (approved):
LGTM! A few comments/questions:

Would highly encourage us to add a note/confirmation before calling AMM stop/close ensuring users their funds will still stay in the safe etc. Also curious: can they reenable it in the future or would they have to create another AMM? Perhaps we could also allow for that. How would the UI show for a user whose AMM has been stopped?

apps/cow-amm-deployer/src/lib/transactionFactory.ts
yvesfracari: On the UI I thought that the `Stop` would be more friendly, but I don't have a strong opinion here actually

yvesfracari:
For now, that is more like a PoC, it will be the same page when there is no AMM running. In the future, I think that we could add a list of all AMMs that the user created and he can enable them (if there isn't any other one activated)

PR #687, opened 2024-05-28
Create Withdraw page

yvesfracari:


apps/cow-amm-deployer/src/components/TokenLogo.tsx
```diff
@@ -36,11 +50,27 @@ export const TokenLogo = ({
   className,
   quality,
 }: ImageFallbackProps) => {
-  const [imageSrc, setImageSrc] = useState<string>();
+  const [imagesSrc, setImagesSrc] = useState<string[]>([
```
ribeirojose, line 53: 🤔 why do we need state here?

yvesfracari: we have a few options of image source for each token, this state should control it.

PR #707, opened 2024-06-18
Pedro/cow 284 milestone review 2

yvesfracari:


apps/cow-amm-deployer/src/contexts/transactionManagerContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  children: React.ReactNode;
+}) => {
+  const managedTransaction = useManagedTransaction();
+
+  return (
+    <TransactionManagerContext.Provider
+      value={{
+        managedTransaction,
+      }}
```
ribeirojose, line 25: any reason why encapsulate this in another object?

apps/cow-amm-deployer/src/contexts/ammDataContext.tsx
```diff
@@ -47,6 +54,19 @@ export const AmmDataContextProvider = ({
...
 
+  useEffect(() => {
+    if (isPonderAPIUpToDate) {
+      mutateAmm();
+      setIsTransactionAwaiting(false);
+    }
+  }, [isPonderAPIUpToDate]);
+
+  useEffect(() => {
+    if (!["final", "confirmed", "error", "idle"].includes(status || "")) {
```
ribeirojose, line 65: we should remove `confirmed` from this now right?

apps/cow-amm-deployer/src/components/DepositForm.tsx
```diff
@@ -153,10 +147,8 @@ export function DepositForm({
         )}
 
       <Button
-        loading={
-          isSubmitting ||
-          !["final", "idle", "confirmed", "error"].includes(status || "")
-        }
+        loadingText="Depositing..."
+        loading={isSubmitting || isAmmUpdating}
```
ribeirojose, line 151: so are we ignoring the tx now 🤔 ?

yvesfracari: actually the `isAmmUpdating` is `true` when the status is something being processed.

apps/cow-amm-deployer/src/contexts/ammDataContext.tsx
yvesfracari: We're using more the ponder update as a trigger, so I think that is ok to let this

